PR #13, opened 2019-06-10
Oleh_Rudko

ghost:


src/App.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.post--dec {
+  color: #B5916E;
+}
+
+.post-list p {
+  width: 600px;
+  margin: 0 auto;
+}
```
gestap:
```suggestion
}

```

src/Components/PostList.jsx
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react';
+import Post from './Post.jsx';
+
+export default class PostList extends Component {
+  render() {
+    return (
+      <div className="post">
+        {this.props.posts.map(post =>
```
gestap:
```suggestion
        {this.props.posts.map(post => (
```

src/Components/PostList.jsx
```diff
@@ -0,0 +1,17 @@
...
+          <Post post={post}
+            key={post.id}
+            users={this.props.users}
+            comments={this.props.comments}
+          />)}
```
gestap:
```suggestion
          />))}
```

src/Components/Comment.jsx
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react'
+
+export default class Comment extends Component {
```
gestap: use function instead of class if you don't use `state`

src/Components/CommentList.jsx
```diff
@@ -0,0 +1,10 @@
+import React, { Component } from 'react'
+import Comment from './Comment.jsx';
+
+export default class CommentList extends Component {
```
gestap: use function instead of class if you don't use `state`

src/Components/Post.jsx
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from 'react'
+import User from './User.jsx';
+import CommentList from './CommentList.jsx';
+
+export default class Post extends Component {
```
gestap: use function instead of class if you don't use `state`

src/Components/PostList.jsx
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react';
+import Post from './Post.jsx';
+
+export default class PostList extends Component {
```
gestap: use function instead of class if you don't use `state`

src/Components/User.jsx
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react'
+
+export default class User extends Component {
```
gestap: use function instead of class if you don't use `state`

PR #19, opened 2019-06-11
Vitaliy Tregubov

VitaliyTregubovKh:


CommentHandler.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'; 
+
+function CommentHandler(props) {
```
YuriiHoliuk: It is not a handler, often handler means some event handler, etc. It is a component, but do not need to name it `CommentComponent`, `Comment` will be enough. Please, rename all handlers

CommentsListHandler.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'; 
+import { comments } from './comments.js';
```
YuriiHoliuk: the only App should know about these files

CommentsListHandler.js
```diff
@@ -0,0 +1,18 @@
...
+import CommentHandler from './CommentHandler.js'
+
+function CommentsListHandler(props){
+  const currentPostComments = comments.filter(comment => comment.postId === props.postId);
+  const commentsList = [];
+  currentPostComments.map(item => {
```
YuriiHoliuk: you should return value in map function,
map callback shouldn`t  have side effects

CommentsListHandler.js
```diff
@@ -0,0 +1,18 @@
...
+    const commentator = item.email;
```
YuriiHoliuk: redundant variable

CommentsListHandler.js
```diff
@@ -0,0 +1,18 @@
...
+    commentsList.push(
+      <CommentHandler commentAuthor={commentator} commentBody={item.body} key={props.key}/>
```
YuriiHoliuk: You haven\`t key in props, also, you shouldn't pass it to the child, use id\`s for keys

PostHandler.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'; 
+import CommentsListHandler from './CommentsListHandler.js';
+
+function PostHandler(props) {
+  return (
+    <div className="post-item">
+      <h2 className="post-title">{props.title}</h2>
```
YuriiHoliuk: will be better to destruct props

PostHandler.js
```diff
@@ -0,0 +1,20 @@
...
+      <p className="post-body">{props.body}</p>
+      <ul className="post-author-info">
+        <li>{props.userName}</li>
+        <li><a href="mailto:{props.userEmail}">{props.userEmail}</a></li>
+        <li>{props.userAddress}</li>
+      </ul>
+      <CommentsListHandler postId={props.postId} key={props.id}/>
```
YuriiHoliuk: do not need to specify key here

PostListHandler.js
```diff
@@ -0,0 +1,33 @@
...
+  const postsList = [];
+  posts.map(post => {
+    const postTitle = post.title;
+    const postBody = post.body;
+    const postUser = users.find(user => user.id === post.userId);
+    const userName = postUser.name;
+    const userAddress = postUser.address.city;
+    const userEmail = postUser.email;
+    const postId = post.id;
+    postsList.push(
```
YuriiHoliuk: do not do side effects in map callback, you should return value

PostListHandler.js
```diff
@@ -0,0 +1,33 @@
...
+function PostsListHandler() {
+  const postsList = [];
+  posts.map(post => {
+    const postTitle = post.title;
+    const postBody = post.body;
+    const postUser = users.find(user => user.id === post.userId);
+    const userName = postUser.name;
+    const userAddress = postUser.address.city;
+    const userEmail = postUser.email;
+    const postId = post.id;
```
YuriiHoliuk: It is ok, but do you really need 7 variables? Also, for better readability, you can use destruction to create them

Comment.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'; 
+
+function Comment(props) {
+  const {
+    commentBody: commentBody,
+    commentAuthor: commentAuthor
```
YuriiHoliuk: you do not need to rename the property to the same name
```javascript
    commentAuthor,
```

CommentsList.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'; 
+import { comments } from './comments.js';
```
YuriiHoliuk: the only App(or PostsList in your case) should know about this file

CommentsList.js
```diff
@@ -0,0 +1,29 @@
...
+  const {
+    postId: currentPostId,
+    key: currentPostKey
+  } = props;
+
+  const currentPostComments = comments.filter(comment => comment.postId === currentPostId);
+
+  const postComments = [];
+
+  currentPostComments.forEach(item => {
```
YuriiHoliuk: use map instead

Post.js
```diff
@@ -0,0 +1,31 @@
...
+
+function Post(props) {
+  const {
+    title: postTitle,
+    body: postBody,
+    userName: postAuthorName,
+    userEmail: postAuthorEmail,
+    userAddress: postAuthorAddress,
+    postId: postId,
+    id: id
```
YuriiHoliuk: duplicated - id's are the same

Post.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'; 
+import CommentsList from './CommentsList.js';
+
+function Post(props) {
+  const {
+    title: postTitle,
+    body: postBody,
+    userName: postAuthorName,
+    userEmail: postAuthorEmail,
+    userAddress: postAuthorAddress,
```
YuriiHoliuk: you do not really need to rename all properties to names with "post", because of you already in Post component

it will be shorter and more readable

PostList.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'; 
+import { posts } from './posts.js';
+import { users } from './users.js';
+import Post from './Post.js';
+
+function PostsList() {
+  const allPosts = [];
+  posts.forEach(post => {
```
YuriiHoliuk: use map instead

Post.js
```diff
@@ -0,0 +1,31 @@
...
+    postId: postId,
+    id: id
+  } = props;
+  return (
+    <div className="post-item">
+      <h2 className="post-title">{postTitle}</h2>
+      <p className="post-body">{postBody}</p>
+      <ul className="post-author-info">
```
YuriiHoliuk: should be separate User component

PostList.js
```diff
@@ -0,0 +1,38 @@
...
+import { users } from './users.js';
+import Post from './Post.js';
+
+function PostsList() {
+  const allPosts = [];
+  posts.forEach(post => { 
+    const {
+      title: postTitle,
+      body: postBody,
+      id: postId
```
YuriiHoliuk: same as above

PostList.js
```diff
@@ -0,0 +1,38 @@
...
+    } = post;
+
+    const postUser = users.find(user => user.id === post.userId);
+    const {
+      name: userName,
+      address: {street, suite, city},
+      email: userEmail
+    } = postUser;
```
YuriiHoliuk: will be better to pass all user here and use it as you want in User component

PR #73, opened 2019-09-09
Solution

ivansafiulin:


src/component/User/User.js
```diff
@@ -0,0 +1,30 @@
...
+
+function User({ user }) {
+  return (
+    <div className="post__author">
+      <address className="author__info">
+        <span className="span__block">
+          <span className="color__text">Name:</span>
+          {user.name}
+        </span>
+        <span className="span__block">
```
irenhh: `span__block` not the best name, choose the one that describes the content

src/component/User/User.js
```diff
@@ -0,0 +1,30 @@
...
+          <span className="color__text">Address:</span>
```
irenhh: the same

src/component/User/User.css
```diff
@@ -0,0 +1,17 @@
+.author__info {
```
irenhh: where is `author` class name?
if there is no, you can't add children to nonexisting parent

PR #84, opened 2019-09-11
Solution

ArtemMarkevych1:
[Link](https://artemmarkevych1.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,29 +1,28 @@
 import React from 'react';
-
 import './App.css';
 
 import posts from './api/posts';
 import comments from './api/comments';
 import users from './api/users';
 
+import PostList from './components/PostList/PostList';
+
+const getPostsWithUserAndComments = (
+  postList,
+  userList,
+  commentList
+) => postList.map(post => ({
+  ...post,
+  user: userList.find(user => user.id === post.userId),
+  comments: commentList.filter(comment => comment.postId === post.id),
+}));
+
+const preparedPosts = getPostsWithUserAndComments(posts, users, comments);
+
 const App = () => (
   <div className="App">
-    <h1>Static list of posts</h1>
-
-    <p>
-      <span>posts: </span>
-      {posts.length}
-    </p>
-
-    <p>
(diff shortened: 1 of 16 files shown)


ArtemMarkevych1:
image
--no-verify because I had troubles with LF and CRLF types of text

src/App.js
```diff
@@ -1,29 +1,28 @@
...
 
-    <p>
-      <span>posts: </span>
-      {posts.length}
-    </p>
+const getPostsWithUserAndComments = (
+  postList,
+  userList,
+  commentList
+) => postList.map(post => ({
```
irenhh: better to open parentheses here and place the function body on the new line

ArtemMarkevych1: Iryna, do you mean like this? Pushed and deployed.
image

src/App.js
```diff
@@ -11,11 +11,13 @@ const getPostsWithUserAndComments = (
   postList,
   userList,
   commentList
-) => postList.map(post => ({
-  ...post,
-  user: userList.find(user => user.id === post.userId),
-  comments: commentList.filter(comment => comment.postId === post.id),
-}));
+) => postList.map(
```
irenhh: I meant this
```
...
) => (
  postList.map(post => ({
    ...post,
    user: userList.find(user => user.id === post.userId),
    comments: commentList.filter(comment => comment.postId === post.id),
  }))
)
```

ArtemMarkevych1: Could you please draw an arrow on your screen, because I cannot understand, where exactly is the point, where the function is going from a new line?

irenhh: `...` represent destructuring and everything before it

ArtemMarkevych1: Like this?
image

irenhh: `postList.map` is the first element of `getPostsWithUserAndComments` function, therefore parentheses should start before it as they wrap the function body

ArtemMarkevych1: Iryna, code is changed and added on github

PR #85, opened 2019-09-11
Max solution

ArnyCorp:


src/components/PostsList/PostsList.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import './PostsList.scss';
+
+import Post from '../Post/Post';
+
+function PostsList({ posts }) {
+  return (
+    <>
```
irenhh: you don't need this since you have one parent element

PR #89, opened 2019-09-11
Solution Andrii

Rafler:


src/App.js
```diff
@@ -1,29 +1,61 @@
...
+      </div>
+      <div className="purple statistic">
+        <div className="value">
+          {comments.length}
+        </div>
+        <div className="label">
+          COMMENTS
+        </div>
+      </div>
+    </div>
```
irenhh: create a separate component for this huge part

Rafler: fixed

PR #146, opened 2020-02-11
solution

MaxLabliuk:


src/Comment.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Comment = ({ oneComment }) => (
+  <div className="comments">
+
```
irenhh: remove unnecessary blank lines

src/CommentList.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Comment from './Comment';
+
+const CommentList = ({ comments }) => (
+  <div>
```
irenhh: do you really need this extra div here?

src/User.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const User = ({ user }) => (
+  <div>
```
irenhh: use React fragment if you don't really need wrapper tags

MaxLabliuk: when i use fragment it ruines my pretty styles)

src/CommentList.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Comment from './Comment';
+
+const CommentList = ({ comments }) => (
+  <React.Fragment>
```
irenhh:
```suggestion
  <>
```

irenhh: short version is enough

PR #147, opened 2020-02-11
Implement static list of posts

AlexanderShpilka:
https://alexandershpilka.github.io/react_static-list-of-posts/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,30 +1,50 @@
 import React from 'react';
-
-import './App.css';
-
 import posts from './api/posts';
 import comments from './api/comments';
 import users from './api/users';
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -1,30 +1,50 @@
...
+const App = () => {
+  const preparedPosts = posts
+    .map((post) => {
+      const user = users.find(it => it.id === post.userId);
+
+      return {
+        ...post,
+        user,
+      };
+    }).map((post) => {
```
irenhh: why don't you do this action in the prev loop?

AlexanderShpilka: my fault.
missed such an opportunity.
fixed.

PR #596, opened 2021-08-31
solution

sasha-morozov:
[DEMO LINK](https://sasha-morozov.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+
+import { CommentInfo } from '../CommentInfo';
+import { Comment } from '../../types/Comment';
+
+import './CommentList.scss';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  return (
+    <div className="comments">
+      <h2 className="comments__title">Comments</h2>
+      <ul className="comments__list">
+        {comments.map(comment => (
+          <li key={comment.id} className="comments__item">
+            <CommentInfo comment={comment} />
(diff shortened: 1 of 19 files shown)


src/components/CommentList/CommentList.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Comment } from '../../types/Comment';
+
+import './CommentList.scss';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  return (
```
DmytroMykoliv, line 13: mentor suggestion is to write like this:
```suggestion

export const CommentList: React.FC<Props> = (props) => {
  const { comments } = props;

```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+type Props = {
+  post: Post;
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => {
+  const {
+    id, title, body, user, commentsLink,
+  } = post;
+
```
DmytroMykoliv, line 17: Same here:
```suggestion
export const PostInfo: React.FC<Props> = (props) => {
 const { post } = props;
  const {
    id, title, body, user, commentsLink,
  } = post;

```

DmytroMykoliv (approved):
Great work, all clear )) 👍

src/components/PostInfo/PostInfo.tsx
sasha-morozov: Thanks for the suggestion!)

danheim (approved):

PR #604, opened 2021-09-01
Develop

ihor-jpeg:
- [DEMO LINK](https://ihor-jpeg.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentInfo/CommentInfo.tsx
+++ b/src/components/CommentInfo/CommentInfo.tsx
@@ -1,0 +1,30 @@
+import { Comment } from '../../types/Comment';
+import './CommentInfo.scss';
+
+type Props = {
+  comment: Comment;
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => {
+  const { name, body, email } = comment;
+
+  return (
+    <>
+      <div className="user-info">
+        <span className="user-info__picture" />
+        <div className="user-info__wrapper">
+          <p className="user-info__title">
+            {name}
+          </p>
+          <p>
+            {email}
+          </p>
+        </div>
+      </div>
+
+      <div className="comments__text">
+        {body}
+      </div>
+    </>
+  );
+};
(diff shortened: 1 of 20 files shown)


AndreySerhovetc (approved):

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -0,0 +1,30 @@
+import { Comment } from '../../types/Comment';
+import './CommentInfo.scss';
+
+type Props = {
+  comment: Comment;
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => {
+  const { name, body, email } = comment;
```
varseniuk, line 9:
```suggestion
export const CommentInfo: React.FC<Props> = (props) => {
  const { comment } = props;
  const { name, body, email } = comment;
```

src/components/CommentList/CommentList.tsx
```diff
@@ -0,0 +1,21 @@
+import { Comment } from '../../types/Comment';
+import { CommentInfo } from '../CommentInfo';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
```
varseniuk, line 8:
```suggestion
export const CommentList: React.FC<Props> = (props) => {
  const { comments } = props;
```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -0,0 +1,32 @@
+import { Post } from '../../types/Post';
+import { CommentList } from '../CommentList';
+import { UserInfo } from '../UserInfo';
+import './PostInfo.scss';
+
+type Props = {
+  post: Post;
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => {
```
varseniuk, line 10:
```suggestion
export const PostInfo: React.FC<Props> = (props) => {
  const { post } = props;
```

varseniuk (approved):
🔥

src/components/PostInfo/PostInfo.tsx
ihor-jpeg: i got this point. is it best practice? i have seen during lectures mentor use object destructuring  when we pass the argument. i do now understand that in that was if we decide to pass more stuff, we won't be able to asses it from props because we have restructured it. thank you!

PR #613, opened 2021-09-02
solution by Eugene Gnedenko

set1one:
[DEMO LINK](https://set1one.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,22 @@
 import React from 'react';
+import { PostList } from './components/PostList';
 
 import './App.scss';
 
-// import posts from './api/posts';
-// import comments from './api/comments';
-// import users from './api/users';
+import posts from './api/posts';
+import comments from './api/comments';
+import users from './api/users';
+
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,14 +1,22 @@
...
-// import posts from './api/posts';
-// import comments from './api/comments';
-// import users from './api/users';
+import posts from './api/posts';
+import comments from './api/comments';
+import users from './api/users';
+
+const preparedPosts = posts.map(post => {
+  const user = users.find(person => person.id === post.userId) || null;
+  const comment = comments.filter(item => item.postId === post.id) || null;
```
ArtemTopolskyi, line 12: it cannot be null
```suggestion
  const comment = comments.filter(item => item.postId === post.id);
```

ArtemTopolskyi (changes requested):
1) use css framework for design
2) add own folder for every component

ArtemTopolskyi (approved):

PR #616, opened 2021-09-03
React_Static_List_Of_Posts

andrii-hryhoruk:
[DEMO LINK](https://andrii-hryhoruk.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -1,1 +1,47 @@
 /// <reference types="react-scripts" />
+
+interface Comment {
+  postId: number;
+  id: number;
+  name: string;
+  email: string;
+  body: string;
+}
+
+interface Geo {
+  lat: string;
+  lng: string;
+}
+
+interface Company {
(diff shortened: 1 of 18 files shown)


src/react-app-env.d.ts
```diff
@@ -1 +1,47 @@
...
+  bs: string;
+}
+
+interface Address {
+  street: string;
+  suite: string;
+  city: string;
+  zipcode: string;
+  geo: Geo;
+}
```
lera-yaskevych, line 28: I think this might be redundant, since you don't display all these props

lera-yaskevych (approved):
Well done! Evererything looks good to me, left just one comment

src/react-app-env.d.ts
andrii-hryhoruk: as Dima said, it's better to describe everything at the beginning

ArtemTopolskyi (approved):

PR #748, opened 2021-12-08
Solution without styling

aldmpr:
[DEMO LINK](https://aldmpr.github.io/react_static-list-of-posts/)


volodymyr-soltys97 (approved):
LGTM 🔥

src/App.tsx
```diff
@@ -2,13 +2,27 @@ import React from 'react';
...
+import comments from './api/comments';
+import users from './api/users';
+
+const preparedPosts = posts.map((post) => {
+  const userLink = users.find(user => post.userId === user.id);
+  const commentsArray = comments.filter(comment => post.id === comment.postId);
+
+  return {
+    user: userLink || null,
+    comments: commentsArray || null,
```
illiarerun, line 17: It'll never be null, cause empty array is still == `true` when converted to boolean
```suggestion
    comments: commentsArray,
```

src/components/PostList/PostList.tsx
```diff
@@ -0,0 +1,29 @@
...
+  posts: Post[]
+};
+
+export const PostList: React.FC<Props> = ({ posts }) => (
+  <ul className="posts">
+    {
+      posts.map(post => (
+        <li key={post.id} className="posts__post">
+          {
+            post.user && (
```
illiarerun, line 16: You're rendering whole post info based on user's value in this post, that's doesn't seem right. Anyway after that you're checking user's value inside this component.
```suggestion
```

illiarerun (changes requested):
Good job 🔥 Few fixes left:

illiarerun (commented):
Also, didn't notice at first, but why do you have all these classes if you have no styles for them?

illiarerun (changes requested):
Remove unnecessary classes or add styles to them

illiarerun (changes requested):
I don't see these styles, you should deploy your solution, use `npm run deploy`

aldmpr:
did npm run deploy but i dont see changes

illiarerun (approved):
GJ

ArtemTopolskyi (approved):

PR #760, opened 2021-12-09
Posts list

WorkAccCom:
[DEMO LINK](https://workacccom.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,7 +22,7 @@ const preparedPosts: PreparedPost[] = posts.map(
 const App: React.FC = () => (
   <div className="App">
     <h1 className="App__title">
-      Post list
+      Posts list
     </h1>
     <PostList readyPosts={preparedPosts} />
   </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,13 +2,29 @@ import React from 'react';
 
 import './App.scss';
 
-// import posts from './api/posts';
-// import comments from './api/comments';
-// import users from './api/users';
+import posts from './api/posts';
+import comments from './api/comments';
+import users from './api/users';
+import { PreparedPost } from './typedefs/PreparedPost';
+import { PostList } from './components/PostList';
+
+const preparedPosts: PreparedPost[] = posts.map(
+  post => {
+    return {
+      ...post,
+
+      user: users.find(user => user.id === post.userId),
+      comments: comments.filter(comment => comment.postId === post.id),
+    };
+  },
+);
 
 const App: React.FC = () => (
   <div className="App">
-    <h1>Static list of posts</h1>
+    <h1 className="App__title">
+      Post list
+    </h1>
+    <PostList readyPosts={preparedPosts} />
   </div>
 );
 
--- a/src/components/CommentInfo/CommentInfo.tsx
+++ b/src/components/CommentInfo/CommentInfo.tsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import { Comment } from '../../typedefs/Comment';
+import '../CommentList/CommentList.scss';
+
+interface Props {
+  singleComment: Comment
+}
+
+export const CommentInfo: React.FC<Props> = ({ singleComment }) => {
+  const {
+    body,
+    name,
+    email,
+  } = singleComment;
+
+  return (
+    <li className="comment-list__list-item">
+      <p className="comment-list__list-item-user-info">
+        <span className="post-list__list-item-name">
+          {name}
+        </span>
+        <br />
+        <a
+          href={`mailto: ${email}`}
+          className="comment-list__list-item-email"
+        >
+          {email}
+        </a>
+      </p>
+      <p className="comment-list__list-item-text">
+        {body}
+      </p>
+    </li>
+  );
+};
--- a/src/components/CommentList/CommentList.scss
+++ b/src/components/CommentList/CommentList.scss
@@ -1,0 +1,16 @@
+.comment-list {
+  list-style: none;
+  padding-left: 0;
+
+  &__list-item {
+    padding: 10px;
+    margin-bottom: 10px;
+
+    background-color: white;
+    border-radius: 10px;
+  }
+
+  &__list-item-email {
+    color: gray;
+  }
+}
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import './CommentList.scss';
+import { CommentInfo } from '../CommentInfo';
+import { Comment } from '../../typedefs/Comment';
+
+interface Props {
+  postComments: Comment[]
+}
+
+export const CommentList: React.FC<Props> = ({ postComments }) => (
+  <ul className="comment-list">
+    {postComments.map(comment => (
+      <CommentInfo singleComment={comment} key={comment.id} />
+    ))}
+  </ul>
+);
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import { PreparedPost } from '../../typedefs/PreparedPost';
+import { UserInfo } from '../UserInfo';
+import { CommentList } from '../CommentList';
+import '../PostList/PostList.scss';
+
+interface Props {
+  singlePost: PreparedPost
+}
+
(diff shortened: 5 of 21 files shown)


src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface Props {
+  singleComment: Comment
+}
+
+export const CommentInfo: React.FC<Props> = ({ singleComment }) => {
+  const {
+    body,
+    name,
+    email,
+  } = singleComment;
```
kurtkobein, line 14:
```suggestion
export const CommentInfo: React.FC<Props> = ({ body, name, email }) => {
...
)
```
I think it's better to pass each prop when we call the component, so we can destructure props fields on arguments

src/typedefs/User.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+  address: Address;
+  phone: string;
+  website: string;
+  company: Company;
+}
```
kurtkobein, line 13:
```suggestion
export interface User {
  name: string,
  email: string,
}
```
You don't need to describe all types of object. Try to describe only fields that you will use.

src/components/CommentList/CommentList.scss
```diff
@@ -0,0 +1,16 @@
...
+    margin-bottom: 10px;
+
+    background-color: white;
+    border-radius: 10px;
+  }
+
+  &__list-item-email {
+    color: gray;
+  }
+}
```
kurtkobein, line 16: blank line

src/components/CommentList/CommentList.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { CommentInfo } from '../CommentInfo';
+import { Comment } from '../../typedefs/Comment';
+
+interface Props {
+  postComments: Comment[]
+}
+
+export const CommentList: React.FC<Props> = ({ postComments }) => (
+  <ul className="comment-list">
+    {postComments.map(comment => (
```
kurtkobein, line 12:
```suggestion
    {postComments.map(({ id, name, body, email }) => (
```

kurtkobein (changes requested):
Great!
Check your interfaces and try to describe only these types of fields that you will use in Components.
Also, try to restructure props on functional component arguments.

src/components/CommentInfo/CommentInfo.tsx
WorkAccCom: This could be better if we would receive only one object as a parameter. But we receive the object as a parameter and restructure it already. This object has another object as a field. And we need properties from this one. How can we restructure right in parameters? In your suggestion, you're destructing the object, but it has no keys _body, _name_, and _email_.

src/components/CommentList/CommentList.tsx
WorkAccCom: why?

src/App.tsx
```diff
@@ -2,13 +2,29 @@ import React from 'react';
...
+const preparedPosts: PreparedPost[] = posts.map(
+  post => {
+    return {
+      ...post,
+
+      user: users.find(user => user.id === post.userId),
+      comments: comments.filter(comment => comment.postId === post.id),
+    };
+  },
+);
```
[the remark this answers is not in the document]
WorkAccCom: I like my way of formating :)

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Comment } from '../../typedefs/Comment';
+import '../CommentList/CommentList.scss';
+
+interface Props {
+  singleComment: Comment
+}
```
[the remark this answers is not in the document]
WorkAccCom: why?

WorkAccCom:
> Great! Check your interfaces and try to describe only these types of fields that you will use in Components. Also, try to restructure props on functional component arguments.

src/typedefs/User.ts
WorkAccCom: Why? Imagine if in the future this app will grow and somewhere I'll need to use more properties from User than these two.

src/typedefs/PreparedPost.ts
```diff
@@ -0,0 +1,12 @@
+import { User } from './User';
+import { Comment } from './Comment';
+
+export interface PreparedPost {
+  userId: number;
+  id: number;
+  title: string;
+  body: string;
+
+  user: User | undefined;
```
illiarerun, line 10:
```suggestion
  user?: User;
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,22 @@
...
+      {userInfo?.name}
+    </span>
+    {' | '}
+    <a
+      href={`mailto: ${userInfo?.email}`}
+      className="post-list__list-item-email"
+    >
+      {userInfo?.email}
+    </a>
+  </p>
```
illiarerun, line 21: Right now UserInfo is heavy dependant on the post-list block, make it so userInfo would be another BEM block

src/components/PostInfo/PostInfo.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <li className="post-list__list-item">
+      <h2 className="post-list__list-item-title">
+        {title}
+      </h2>
+      <p className="post-list__list-item-text">
+        {body}
+      </p>
+
+      <UserInfo userInfo={user} />
```
illiarerun, line 28: Your user prop might be undefined so use conditional rendering here, `&&` operator will do it
```suggestion
      {user && <UserInfo userInfo={user} />}
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { User } from '../../typedefs/User';
+import '../PostList/PostList.scss';
+
+interface Props {
+  userInfo: Pick<User, 'email' | 'name'> | undefined,
```
illiarerun, line 6: Then you wouldn't need this undefined type here
```suggestion
  userInfo: Pick<User, 'email' | 'name'>,
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export const UserInfo: React.FC<Props> = ({ userInfo }) => (
+  <p className="post-list__list-item-user-info">
+    <span className="post-list__list-item-name">
+      {userInfo?.name}
```
illiarerun, line 12: And this question marks too
```suggestion
      {userInfo.name}
```

illiarerun (changes requested):
GJ 🔥 Let's fix few things now:

src/components/CommentInfo/CommentInfo.tsx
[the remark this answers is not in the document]
kurtkobein: Check this [video ](https://www.youtube.com/watch?v=YEJ4OboyvF4), timecode 5:25

src/typedefs/User.ts
kurtkobein: So you can add these properties to your interface

src/components/CommentInfo/CommentInfo.tsx
kurtkobein: Yeah, so you need to pass each prop when you call CommentInfo from CommentList component. 
Check this [video](https://www.youtube.com/watch?v=LGqqEg1Z4FA), timecode - 2:58

src/components/CommentList/CommentList.tsx
kurtkobein: Cause we can see fields that comment include and we can pass each of them to CommentInfo component.

src/components/UserInfo/UserInfo.tsx
WorkAccCom: I do not understand, why?

src/components/CommentList/CommentList.tsx
ihor-jpeg: if we want to see what fields a particular type or an interface includes, we can check the interface itself. destructuring unnecessary props here would be similar to declaring a variable and not using it

ihor-jpeg (commented):
Good work!

ihor-jpeg (approved):

PR #807, opened 2022-01-19
solution

Rom911:
[demo](https://rom911.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -18,6 +18,9 @@ interface Post {
   id: number;
   title: string;
   body: string;
+}
+
+interface PreparedPost extends Post {
   user: User | null;
   comments: CommentItem[];
 }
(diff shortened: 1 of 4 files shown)


src/react-app-env.d.ts
```diff
@@ -1 +1,23 @@
...
+}
+
+interface Post {
+  userId: number;
+  id: number;
+  title: string;
+  body: string;
+  user: User | null;
+  comments: CommentItem[];
+}
```
Rozdorozhnii, line 23: You can also inherit some instance and use union interface where need it
`interface Post {
  userId: number;
  id: number;
  title: string;
  body: string;
}
interface prepearedPost extend Post {
  user: User | null;
  comments: CommentItem[];
}
`

Rozdorozhnii (approved):

Elena-brv (approved):
🔥

PR #888, opened 2022-04-17
Result

xpromt:
https://xpromt.github.io/react_static-list-of-posts/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # React static list of posts
 - Replace `<your_account>` with your Github username in the
-  [DEMO LINK](https://<your_account>.github.io/react_static-list-of-posts/)
+  [DEMO LINK](https://xpromt.github.io/react_static-list-of-posts/)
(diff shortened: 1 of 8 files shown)


Bohdan-Maksymiuk (changes requested):
Same. Looks good, but add all new files to git and then I could check it
`git add .` before committing

xpromt:
Done

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -0,0 +1,37 @@
...
+    >
+      {comment.name}
+    </h3>
+
+    <span
+      className="comment-info__user-info"
+      data-cy="comment-email"
+    >
+      <i className="fa fa-envelope" />
+      {' '}
```
ArtemTopolskyi, line 23: use template string

ArtemTopolskyi (approved):

PR #946, opened 2022-05-29
Develop

alexandra-protyanova:


khozhainov-aleksandr (approved):
Everything is fine 👍🏻  Вut the HTML is almost all on the `p` tags, not very good! Better more symantics 🥳  And the styles are very bright colors 😵‍💫

src/App.tsx
```diff
@@ -1,14 +1,38 @@
...
+import comments from './api/comments';
+import users from './api/users';
+import {
+  Post,
+  User,
+  Comment,
+  FullPost,
+} from './react-app-env';
+
+const preparingPosts = (
```
Anna-Tkachenko, line 16:
```suggestion
const getPreparedPosts = (
```

Anna-Tkachenko (approved):
⭐️

PR #948, opened 2022-05-29
Task Done v1.0

khozhainov-aleksandr:
[DEMO LINK](https://khozhainov-aleksandr.github.io/react_static-list-of-posts/)


alexandra-protyanova (approved):
Good job)

src/App.tsx
```diff
@@ -1,15 +1,49 @@
...
+    postsList: PostsFromServer[],
+    commentsList: CommentsFromServer[],
+    usersList: UsersFromServer[],
+  ): GeneralPosts[] => {
+    return postsList.map(post => ({
+      ...post,
+      user: usersList
+        .find(user => post.userId === user.id)
+        || null,
+      comments: commentsList
```
Anna-Tkachenko, line 26: Method filter always returns an array, so null never be the value of comments field

src/components/PostInfo/PostInfo.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+      <p className="postsListItemBody" data-cy="post-body">
+        {body}
+      </p>
+
+      <div className="postsListItemUserInfo">
+        {user && (
+          <UserInfo mainUser={user} />
+        )}
+      </div>
```
Anna-Tkachenko, line 33: Don't render this div if there is no user
{user && (
  <div> ... </div>
)}

src/components/PostInfo/PostInfo.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+      <div className="postsListItemComments">
+        {comments && (<CommentList commentsList={comments} />)}
+      </div>
```
Anna-Tkachenko, line 37: Same here with comments
Comments always contain some array, so this check comments && is redundant

src/react-app-env.d.ts
```diff
@@ -1 +1,45 @@
 /// <reference types="react-scripts" />
+
+export interface PostsFromServer {
```
Anna-Tkachenko, line 3: It's type for one post
```suggestion
export interface PostFromServer {
```

src/react-app-env.d.ts
```diff
@@ -1 +1,45 @@
...
+  userId: number,
+  id: number,
+  title: string;
+  body: string;
+}
+
+export interface CommentsFromServer {
```
Anna-Tkachenko, line 10: It's type for one comment
```suggestion
export interface CommentFromServer {
```

src/react-app-env.d.ts
```diff
@@ -1 +1,45 @@
...
+  postId: number;
+  id: number;
+  name: string;
+  email: string;
+  body: string;
+}
+
+export interface UsersFromServer {
```
Anna-Tkachenko, line 18: It's type for one user
```suggestion
export interface UserFromServer {
```

src/react-app-env.d.ts
```diff
@@ -1 +1,45 @@
...
+  company: {
+    name: string;
+    catchPhrase: string;
+    bs: string;
+  },
+}
+
+export interface GeneralPosts extends PostsFromServer {
+  user: UsersFromServer | null;
+  comments: CommentsFromServer[] | null;
```
Anna-Tkachenko, line 44: comments always store some array from filter method

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #954, opened 2022-05-31
Posts

anastasiia-nikita:
[DEMO LINK](https://anastasiia-nikita.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,13 +2,41 @@ import React from 'react';
 
 import './App.scss';
 
-// import posts from './api/posts';
-// import comments from './api/comments';
-// import users from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import {
+  Post,
+  Comment,
+  User,
+  PreparedPosts,
+} from './react-app-env';
+
+import { PostList } from './components/PostList/PostList';
+
+const prepare = (
+  posts: Post[],
+  comments: Comment[],
+  users: User[],
+): PreparedPosts[] => {
+  return posts.map(post => ({
+    ...post,
+    user: users.find(user => post.userId === user.id),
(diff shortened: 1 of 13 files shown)


src/components/CommentList/CommentList.tsx
```diff
@@ -1,6 +1,24 @@
 import React from 'react';
 import './CommentList.scss';
+import { Comment } from '../../react-app-env';
+import { CommentInfo } from '../CommentInfo/CommentInfo';
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
+interface Props {
```
dellachaise, line 6: I think you should replace interface with type. Because in our videos said, that interface we use for describing data, and type for our inner needs(for example Props).

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,6 +1,24 @@
 import React from 'react';
+import { PreparedPosts } from '../../react-app-env';
+import { CommentList } from '../CommentList/CommentList';
+import { UserInfo } from '../UserInfo/UserInfo';
 import './PostInfo.scss';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+interface Props {
```
dellachaise, line 7: The same way as previous comment - please replace interface with type.

dellachaise (approved):

src/App.tsx
```diff
@@ -2,13 +2,41 @@ import React from 'react';
...
+import {
+  Post,
+  Comment,
+  User,
+  PreparedPosts,
+} from './react-app-env';
+
+import { PostList } from './components/PostList/PostList';
+
+const prepare = (
```
Anna-Tkachenko, line 17:
```suggestion
const getPreparedPosts = (
```

Anna-Tkachenko (approved):
⭐️

PR #967, opened 2022-06-01
posts

Oleksandr-Ivanchenko:
[DEMO LINK](https://Oleksandr-Ivanchenko.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,13 +2,38 @@ import React from 'react';
 
 import './App.scss';
 
-// import posts from './api/posts';
-// import comments from './api/comments';
-// import users from './api/users';
+import postsFromServer from './api/posts';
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -2,13 +2,38 @@ import React from 'react';
...
+const preparePosts = (
+  comments: Comment[],
+  users: User[],
+  posts: Post[],
+) => {
+  return posts.map(post => ({
+    ...post,
+    user: users.find(user => post.userId === user.id) || null,
+    comments: comments
+      .filter(comment => post.id === comment.postId) || null,
```
Anna-Tkachenko, line 24: Method filter always return an array, so comments will never get null value

Anna-Tkachenko (approved):

PR #977, opened 2022-06-03
Develop

MabuzaM:
[DEMO LINK](https://MabuzaM.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,23 +21,22 @@ export function preparePosts(
     ...post,
     user: users.find(user => (
       user.id === post.userId
-    )) || null || undefined,
+    )) || null,
     postComments: postComments.filter(postComment => (
       postComment.postId === post.id
-    )) || null || undefined,
+    )) || null,
   }));
 }
-
-export const preparedPosts = preparePosts(
-  postsFromServer,
-  commentsFromServer,
-  usersFromServer,
-);
 
 const App: React.FC = () => (
   <div className="App">
     <h1>Static list of posts</h1>
-    <PostList posts={preparedPosts} />
+    <PostList posts={preparePosts(
+      postsFromServer,
+      commentsFromServer,
+      usersFromServer,
+    )}
+    />
   </div>
 );
 
--- a/src/appTypeDefs.ts
+++ b/src/appTypeDefs.ts
@@ -45,5 +45,5 @@ type Company = {
 
 export interface PreparedPost extends Post {
   postComments: PostComment[];
-  user?: User;
+  user: User | null;
 }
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -1,14 +1,43 @@
...
+export function preparePosts(
+  posts: Post[],
+  postComments: PostComment[],
+  users: User[],
+): PreparedPost[] {
+  return posts.map(post => ({
+    ...post,
+    user: users.find(user => (
+      user.id === post.userId
+    )) || null || undefined,
```
iLebo-stack, line 24: Why do you include `undefined`?

src/App.tsx
```diff
@@ -1,14 +1,43 @@
...
+      postComment.postId === post.id
+    )) || null || undefined,
+  }));
+}
+
+export const preparedPosts = preparePosts(
+  postsFromServer,
+  commentsFromServer,
+  usersFromServer,
+);
```
iLebo-stack, line 35: I'm not sure about the helper function I don't think you really need it since you only use it once. I think you could've assigned the body of the return statement to `preparedPosts`

src/appTypeDefs.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type Company = {
+  name: string;
+  catchPhrase: string;
+  bs: string;
+};
+
+export interface PreparedPost extends Post {
+  postComments: PostComment[];
+  user?: User;
```
iLebo-stack, line 48: `user` is not optional. It should be there but if it's value is falsy it should be `null`. Just like you assigned `null` if user is not found by `user.id` in your helper function.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,7 +1,14 @@
 import React from 'react';
-
 import './UserInfo.scss';
+import { User } from '../../appTypeDefs';
+
+interface UserInfoProps {
+  user: User | null | undefined;
```
iLebo-stack, line 6: Why `undefined`?

iLebo-stack (changes requested):
Good job 👍. Check my comments

src/components/UserInfo/UserInfo.tsx
MabuzaM: Removed :) Thanx

src/appTypeDefs.ts
MabuzaM: Fixed it thank you.

src/App.tsx
MabuzaM: Fixed that thank you

src/App.tsx
MabuzaM: Okay yah you are right. Have fixed it. Thanx

PR #987, opened 2022-06-06
solution

Uktam19980416:
[DEMO LINK](https://Uktam19980416.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,6 +1,29 @@
 import React from 'react';
 import './PostInfo.scss';
+import { PreparedPosts } from '../../types';
+import { UserInfo } from '../UserInfo';
+import { CommentList } from '../CommentList';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+interface Props {
+  post: PreparedPosts;
+}
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <>
+    <div className="post" data-cy="post-info">
+      <h2 className="post__title" data-cy="post-title">{post.title}</h2>
+      <p className="post__body" data-cy="post-body">{post.body}</p>
+
+      <div className="post__user">
+        {post.user && (
+          <UserInfo user={post.user} />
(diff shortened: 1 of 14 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,6 +1,29 @@
...
+import { CommentList } from '../CommentList';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+interface Props {
+  post: PreparedPosts;
+}
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <>
```
59MrRobot, line 12: You don't need to use fragments here because the div already returns a single block

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,6 +1,29 @@
...
+
+      <div className="post__user">
+        {post.user && (
+          <UserInfo user={post.user} />
+        )}
+      </div>
+
+      <div className="post__comments">
+        <h4 className="post__comments-title">Comments:</h4>
+        <CommentList comments={post.comments} />
```
59MrRobot, line 25: I think you should destructure the object so you don't need to continuously write `post.`

59MrRobot (changes requested):
Good job, just a couple suggestions

src/components/PostInfo/PostInfo.tsx
```diff
@@ -8,11 +8,16 @@ interface Props {
...
+
+  return (
     <div className="post" data-cy="post-info">
-      <h2 className="post__title" data-cy="post-title">{post.title}</h2>
-      <p className="post__body" data-cy="post-body">{post.body}</p>
+      <h2 className="post__title" data-cy="post-title">{title}</h2>
+      <p className="post__body" data-cy="post-body">{body}</p>
 
       <div className="post__user">
         {post.user && (
```
meforoma, line 23: since you're already de-structuring - why not de-structure `user, comments` ?

meforoma (approved):

PR #991, opened 2022-06-06
code is like in the previous task

WojciechCzarnowski:
[DEMO LINK](https://WojciechCzarnowski.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,6 +1,20 @@
 import React from 'react';
 import './CommentList.scss';
+import { Comment } from '../../types';
+import { CommentInfo } from '../CommentInfo/CommentInfo';
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
+type Props = {
+  comment: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comment }) => (
+  <>
+    <ul className="comment-list">
+      {comment.map(comms => (
+        <li key={comms.id}>
+          <CommentInfo comment={comms} />
+        </li>
+      ))}
+    </ul>
+  </>
 );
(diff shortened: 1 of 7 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,6 +1,19 @@
...
+type Props = {
+  post: ExtendedPost;
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <div data-cy="post-info" className="post-info">
+    <h2 data-cy="post title" className="post-title">{post.title}</h2>
+    <p data-cy="post body" className="post-body">{post.body}</p>
+    {post.user && <UserInfo user={post.user} />}
+    <h3 className="post-comments">{post.comments}</h3>
```
meforoma, line 16: @WojciechCzarnowski `post.comments` is an array, right?
this line tries to render an array, which is not possible.
You do the proper rendering in the next line by passing `post.comments` to `<CommentList` component
```suggestion
```

src/components/CommentList/CommentList.tsx
```diff
@@ -1,6 +1,20 @@
...
+import { CommentInfo } from '../CommentInfo/CommentInfo';
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
+type Props = {
+  comment: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comment }) => (
+  <>
```
meforoma, line 11: no need for a fragment here
<ul> does the wrapping - so your JSX returns one html element
```suggestion
```

meforoma: ofc remove the closing tag as well

meforoma (changes requested):
try addressing the comments
and re-deploy, - should work

meforoma (approved):

PR #1023, opened 2022-07-07
Solution

sancho325:
[DEMO LINK](https://sancho325.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,25 @@
 import React from 'react';
 
 import './App.scss';
+import { PostList } from './components/PostList/PostList';
 
-// import posts from './api/posts';
-// import comments from './api/comments';
-// import users from './api/users';
+import posts from './api/posts';
+import comments from './api/comments';
+import users from './api/users';
+
+const preparedPosts = posts.map(post => (
+  {
+    ...post,
+    user: users.find(user => user.id === post.userId) as User,
+    comments: comments
+      .filter(comment => comment.postId === post.id) as Comment[],
+  }
+));
 
 const App: React.FC = () => (
   <div className="App">
     <h1>Static list of posts</h1>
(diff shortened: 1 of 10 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,6 +1,36 @@
...
+      <div className="post__flex">
+        <h1 className="post__title" data-cy="post-title">
+          {title}
+        </h1>
+
+        <p className="post__body" data-cy="post-body">
+          {body}
+        </p>
+
+        <UserInfo user={user} />
```
Olexandr6, line 30: Should we render this component if `user === null`? Let's use conditional rendring

src/App.tsx
```diff
@@ -1,14 +1,25 @@
...
-// import comments from './api/comments';
-// import users from './api/users';
+import posts from './api/posts';
+import comments from './api/comments';
+import users from './api/users';
+
+const preparedPosts = posts.map(post => (
+  {
+    ...post,
+    user: users.find(user => user.id === post.userId) as User,
```
Olexandr6, line 13: `find` returns `User` or `undefined`, so there can be cases when user is undefined it's not okay

Olexandr6 (changes requested):
Almost done👍 Remember, you should avoid using typecasting while use TS

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,6 +1,36 @@
...
+  post: Post,
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => {
+  const {
+    title,
+    body,
+    user,
+    comments,
+  } = post;
```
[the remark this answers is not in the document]
sancho325: 🔥

illiarerun (approved):

PR #1049, opened 2022-07-11
Develop

DmitriiPirohov:
[DEMO LINK](https://DmitriiPirohov.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,3 +1,11 @@
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
+import React from 'react';
+import { User } from '../../type';
+
+import './UserInfo.scss';
+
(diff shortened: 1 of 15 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,3 +1,11 @@
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
+import React from 'react';
+import { User } from '../../type';
+
+import './UserInfo.scss';
+
+export const UserInfo: React.FC<{ user: User }> = ({ user }) => (
```
Olexandr6, line 6: Let's follow this approach, it's easily to read
interface Props {
  ...
}
```suggestion
export const UserInfo: React.FC<Props> = ({ user }) => (
```

Olexandr6 (changes requested):
DEMO link doesn't work, check if you deployed your solution

PR #1058, opened 2022-07-11
Static list of posts

garikr:
[DEMO LINK](https://garikr.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,3 +1,31 @@
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+import React from 'react';
+import { Post } from '../../types/Post';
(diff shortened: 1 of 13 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,3 +1,31 @@
...
+type Props = {
+  post: Post
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <div className="PostInfo">
+    <div className="PostInfo__header">
+      <h3 className="PostInfo__title">{post.title}</h3>
+
+      {' Posted by  '}
```
Olexandr6, line 16: This test should be moved to `UserInfo` component, cause in case user === null, this text will look strange

Olexandr6 (approved):
Good solution🔥

PR #1137, opened 2022-08-31
solution

tiserett:
[DEMO LINK](https://tiserett.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,24 +3,7 @@ import React from 'react';
 import './App.scss';
 
 import { PostList } from './components/PostList';
-import { Post } from './types/Post';
-import { Comment } from './types/Comment';
-
-import postsFromServer from './api/posts';
-import commentsFromServer from './api/comments';
-import usersFromServer from './api/users';
-
-function getComments(postId: number): Comment[] {
-  return commentsFromServer.filter(comment => (
-    comment.postId === postId
-  ));
-}
-
-const posts: Post[] = postsFromServer.map(post => ({
-  ...post,
-  comments: getComments(post.id),
-  user: usersFromServer.find(user => user.id === post.userId) || null,
-}));
+import { posts } from './helpers/generatePosts';
 
 export const App: React.FC = () => (
   <section className="App">
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,32 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import { PostList } from './components/PostList';
+import { Post } from './types/Post';
+import { Comment } from './types/Comment';
+
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+function getComments(postId: number): Comment[] | null {
+  const comments = commentsFromServer.filter(comment => (
+    comment.postId === postId
+  ));
+
+  return comments.length > 0 ? comments : null;
+}
+
+const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  comments: getComments(post.id),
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -2,108 +2,32 @@ import React from 'react';
...
+    comment.postId === postId
+  ));
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+  return comments.length > 0 ? comments : null;
+}
```
sergii-nosachenko, line 19:
```suggestion
function getComments(postId: number): Comment[] {
  return commentsFromServer.filter(comment => (
    comment.postId === postId
  ));
}
```

I have doubts about this function work. You've decided to return null if no posts found. But this decision created a new type checking branch. What if I propose you to return only one type of data - `Comment[]`. Array filter method in any case will return you an array, so you don't need to worry about data type change.

So you can remove null from everywhere and check only if `comments.length` convertes to `true`.

What do you think?

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{ title }</h3>
+
+        <p>
+          {' Posted by  '}
+
+          {user
+            ? <UserInfo user={user} />
+            : ''}
+        </p>
```
sergii-nosachenko, line 30:
```suggestion
{user && <UserInfo user={user} />}
```

1. This paragraph belongs to UserInfo (if no user - no paragraph with text Posted by...)
2. Same condition can be achieved with AND operator.

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+      </div>
+
+      <p className="PostInfo__body">
+        { body }
+      </p>
+
+      {comments
```
sergii-nosachenko, line 37:
```suggestion
      {comments.length
```

If you return `Comment[]`

src/types/Post.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from './User';
+import { Comment } from './Comment';
+
+export interface Post {
+  userId: number,
+  id: number,
+  title: string,
+  body: string,
+  comments: Comment[] | null,
```
sergii-nosachenko, line 9:
```suggestion
  comments: Comment[],
```

sergii-nosachenko (changes requested):
I have found some controversial solutions it your code, that need to be discussed.

Also I've noticed that you have left styles untouched, but you need to separate'em as you did it it with components (cause they are separate BEM blocks).

src/App.tsx
tiserett: sounds like a way to get rid of unnecessary written code

sergii-nosachenko (approved):
Perfect 🔥

src/App.tsx
```diff
@@ -2,108 +2,30 @@ import React from 'react';
...
+  return commentsFromServer.filter(comment => (
+    comment.postId === postId
+  ));
+}
+
+const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  comments: getComments(post.id),
+  user: usersFromServer.find(user => user.id === post.userId) || null,
+}));
```
AlinaMatuschak, line 23: you can create folder `helpers` with file `generatePosts` with this function

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,36 @@
...
+        {user && <UserInfo user={user} />}
+      </div>
+
+      <p className="PostInfo__body">
+        { body }
+      </p>
+
+      {comments.length
+        ? <CommentList comments={comments} />
+        : <b data-cy="NoCommentsMessage">No comments yet</b>}
```
AlinaMatuschak, line 33: ❤️

AlinaMatuschak (approved):

PR #1215, opened 2022-09-25
Task done

sanyokmalyshev:
[DEMO LINK](https://sanyokmalyshev.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/types/Post.ts
+++ b/src/types/Post.ts
@@ -1,0 +1,11 @@
+import { Comment } from './Comment';
+import { User } from './User';
+
+export interface Post {
+  userId: number,
+  id: number,
+  title: string,
+  body: string,
+  comments: Comment[] | [],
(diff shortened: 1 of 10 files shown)


src/types/Post.ts
```diff
@@ -0,0 +1,11 @@
+import { Comment } from './Comment';
+import { User } from './User';
+
+export interface Post {
+  userId: number,
+  id: number,
+  title: string,
+  body: string,
+  comments: Comment[] | [],
```
qoa11a, line 9: You can write just Comment[] as an empty array is also Comment[]

qoa11a (changes requested):

src/types/Post.ts
sanyokmalyshev: Done

PR #1232, opened 2022-10-10
Develop

grafit-off:
[DEMO LINK](https://grafit-off.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/api/getComments.ts
+++ b/src/api/getComments.ts
@@ -1,0 +1,9 @@
+import comments from './comments';
+
+import { Comment } from '../types/Comment';
+
+type GetComments = (postId: number) => Comment[];
+
+export const getComments: GetComments = (postId) => (
+  comments.filter(comment => comment.postId === postId)
+);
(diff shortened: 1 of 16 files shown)


vmamchur (approved):
Looks good to me 👍🏻

src/api/getComments.ts
```diff
@@ -0,0 +1,9 @@
+import comments from './comments';
+
+import { Comment } from '../types/Comment';
```
naziamoff, line 3: No need to use empty line here
```suggestion
import comments from './comments';

import { Comment } from '../types/Comment';
```

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,35 @@
...
+
+  return (
+    <div className="CommentInfo">
+      <div className="CommentInfo__title is-flex is-align-items-center">
+        <strong className="CommentInfo__name mr-1">
+          {name}
+        </strong>
+
+        by
+
```
naziamoff, line 22: do you need those empty lines here?

naziamoff (approved):
🔥

src/components/CommentInfo/CommentInfo.tsx
grafit-off: I forgot to remove this one, but to be honest I like to visually decompose imports that have different purposes, but when we have only two imports maybe it's looks redundant

PR #1254, opened 2022-10-13
Solution

Yeliseiev:
[DEMO LINK](https://Yeliseiev.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,35 @@
 import React from 'react';
-
+import { PostList } from './components/PostList';
 import './App.scss';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { Post } from './types/Post';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { FullPost } from './types/FullPost';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+const preparePosts = (
+  posts: Post[],
+  users: User[],
+  comments: Comment[],
+): FullPost[] => {
+  return posts.map(post => ({
+    ...post,
+    user: users.find(user => user.id === post.userId) || null,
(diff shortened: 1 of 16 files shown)


Maksym845 (approved):
I didn`t find any mistakes, well done!👍

src/App.tsx
```diff
@@ -1,109 +1,35 @@
...
+
+const preparePosts = (
+  posts: Post[],
+  users: User[],
+  comments: Comment[],
+): FullPost[] => {
+  return posts.map(post => ({
+    ...post,
+    user: users.find(user => user.id === post.userId) || null,
+    comments: comments.filter(comment => comment.postId === post.id) || [],
```
OleksiiBlezniuk, line 20: filter returns empty array if there are no elements that satisfied condition

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,16 @@
...
-  <>Put the list here</>
+type Props = {
+  comments: Comment[],
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <div className="CommentList">
+    {comments.map(comment => (
+      <CommentInfo comment={comment} key={comment.id} />
+    ))}
```
OleksiiBlezniuk, line 14: better to use semantic tags for lists

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,17 @@
...
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <p>
+    {` Posted by  ${user?.username}: `}
+
+    <a className="UserInfo" href={`mailto:${user?.email}`}>
+      {user?.name}
+    </a>
+  </p>
```
OleksiiBlezniuk, line 16: And what will be here if there is no user?
Just don't render this component if user is null (check it in parent component)

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #1271, opened 2022-10-14
list of posts was rendered

iPanda177:
[DEMO LINK](https://iPanda177.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -6,6 +6,10 @@ iframe {
   text-align: center;
 }
 
+.PostList {
+  list-style: none;
+}
+
 .PostInfo {
   margin: 10px auto;
   width: 90%;
@@ -33,4 +37,6 @@ iframe {
   gap: 0.7em;
   background-color: #eee;
   padding: 1em;
+
+  list-style: none;
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,22 @@
 import React from 'react';
-
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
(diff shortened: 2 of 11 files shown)


src/App.scss
```diff
@@ -6,6 +6,10 @@ iframe {
   text-align: center;
 }
 
+.PostList {
```
iam-datsenko, line 9: Move each one to the appropriate file (PostList to the PostList.scss and soon)

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,26 @@
...
-  <>Put the comment here</>
+type Props = {
+  comment: Comment,
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => (
+  <li className="CommentInfo">
+    <div className="CommentInfo__title">
+      <strong className="CommentInfo__name">
+        {comment.name}
```
iam-datsenko, line 12: Use destructurization everywhere

iam-datsenko (changes requested):

src/App.tsx
```diff
@@ -1,109 +1,22 @@
...
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+import { PostList } from './components/PostList';
+
+const configuredPosts = postsFromServer.map(post => ({
+  ...post,
+  user: usersFromServer.find(user => user.id === post.userId),
```
Anna-Tkachenko, line 12: will be better to set null if there is no user in usersFromServer array
in this way you can be sure that you definitely set null in this prop, and it's not undefined value because of you don't have this prop

Anna-Tkachenko (approved):
Good job!

PR #1272, opened 2022-10-14
list of posts was created

vallerii:


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,39 @@
 import React from 'react';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+import './PostInfo.scss';
+
+import { Post } from '../../types/Post';
+import { UserInfo } from '../UserInfo';
+import { CommentList } from '../CommentList';
+import comments from '../../api/comments';
```
korniienkokarina, line 8: you can delete 8th line, so linter will not trow an error, as you use comments from post, not directly from api

korniienkokarina (commented):
[demo link](https://vallerii.github.io/react_static-list-of-posts/) is working and everything looks good, left a comment about linter 😊

src/App.tsx
```diff
@@ -1,109 +1,41 @@
...
-            {' Posted by  '}
+  return foundUser || null;
+}
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+function getComment(postId: number): Comment[] | [] {
```
naziamoff, line 20: Better to explicitly mention the way we going to get the comment. Not critical in that case imho
```suggestion
function getCommentById(postId: number): Comment[] | [] {
```

naziamoff (changes requested):

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,22 @@
...
+
+import { Comment } from '../../types/Coment';
+import { CommentInfo } from '../CommentInfo';
+
+type Props = {
+  comments: Comment[]
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  const { id } =comments;
```
naziamoff, line 13: no no no 😄 
Screenshot 2022-10-25 at 11 05 41

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,22 @@
...
+
+  return (
+    <div className="CommentList">
+      {comments.map(comment => (
+        <CommentInfo comment={comment} key={id} />
+      ))}
+    </div>
```
naziamoff, line 20: and better to use semantics for rendering a list

naziamoff (changes requested):

src/components/CommentList/CommentList.tsx
vallerii: oh, my bad

src/App.tsx
```diff
@@ -1,109 +1,41 @@
...
+function getCommentById(postId: number): Comment[] | [] {
+  const foundComment = commentsFromServer
+    .filter(comment => comment.postId === postId);
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+  return foundComment || [];
```
Anna-Tkachenko, line 24: method filter return an empty array if there is no elements which pass the check
so it's redundant to return an empty array here one more time, this check doesn't work

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,19 @@
 import React from 'react';
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
+import './UserInfo.scss';
+
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null
```
Anna-Tkachenko, line 8: you render this component only if user exists, so you shouldn't handle null value here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1281, opened 2022-10-16
Solution

rstarovoit:
[DEMO LINK](https://rstarovoit.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,5 +1,17 @@
 import React from 'react';
+import { CommentInfo } from '../CommentInfo';
+import { Comment } from '../../react-app-env';
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
+type Props = {
+  comments: Comment[],
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <div className="CommentList">
+    {
+      comments.map(comment => (
+        <CommentInfo key={comment.id} comment={comment} />
+      ))
+    }
+  </div>
 );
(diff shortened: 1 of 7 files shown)


src/react-app-env.d.ts
```diff
@@ -1 +1,25 @@
...
+}
+
+export interface Post {
+  userId: number;
+  id: number;
+  title: string;
+  body: string;
+  user: User | null;
+  comments: Comment[];
+}
```
AndriyZhabynets, line 25: I think it would be better to create a separate file for each interface

AndriyZhabynets (approved):
GJ!

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,17 @@
...
+  comments: Comment[],
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <div className="CommentList">
+    {
+      comments.map(comment => (
+        <CommentInfo key={comment.id} comment={comment} />
+      ))
+    }
```
OleksiiBlezniuk, line 15: use semantic tags for list and its items

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,38 @@
...
+  const userToFind = usersFromServer.find(user => user.id === userId);
+
+  return userToFind || null;
+}
+
+function getComments(postId: number): Comment[] {
+  const comments = commentsFromServer
+    .filter(comment => postId === comment.postId);
+
+  return comments;
```
OleksiiBlezniuk, line 20:
```suggestion
  return commentsFromServer
    .filter(comment => postId === comment.postId);
```

OleksiiBlezniuk: Ok, but no need to save result in variable in this case, we can just return it

OleksiiBlezniuk (approved):

PR #1285, opened 2022-10-16
v1: list of posts has been completed

igor-1987:
[DEMO LINK](https://igor-1987.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,23 @@
 import React from 'react';
-
+import { Post } from './react-app-env';
+import { PostList } from './components/PostList';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: usersFromServer.find(user => user.id === post.userId),
+  comments: commentsFromServer.filter(comment => comment.postId === post.id),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
 
-    <div className="PostList">
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">qui est esse</h3>
-
-          <p>
-            {' Posted by  '}
-
-            <a className="UserInfo" href="mailto:Sincere@april.biz">
-              Leanne Graham
-            </a>
-          </p>
-        </div>
-
-        <p className="PostInfo__body">
-          est rerum tempore vitae sequi sint nihil reprehenderit dolor beatae
-          ea dolores neque fugiat blanditiis voluptate porro vel nihil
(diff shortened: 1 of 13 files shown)


poznianski (approved):

src/App.tsx
```diff
@@ -1,109 +1,23 @@
...
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: usersFromServer.find(user => user.id === post.userId),
```
Anna-Tkachenko, line 11: will be better to set null here if user wasn't found

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,24 @@
...
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
-);
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  const commentsList = comments.map(comment => (
```
Anna-Tkachenko, line 11: move map logic inside the return statement, don't create separate variable for this

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,47 @@
...
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{title}</h3>
+
+        <p>
+          {' Posted by  '}
+          {user && <UserInfo user={user} />}
+        </p>
```
Anna-Tkachenko, line 27: don't show this paragraph if there is no user

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,109 +1,24 @@
...
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+const posts: Post[] | [] = postsFromServer.map(post => ({
+  ...post,
+  user: usersFromServer.find(user => user.id === post.userId) || null,
+  comments: commentsFromServer
+    .filter(comment => comment.postId === post.id) || [],
```
Anna-Tkachenko, line 13: filter method returns empty array by default if it's not found any elements, || [] here is redundant

src/react-app-env.d.ts
```diff
@@ -1 +1,25 @@
 /// <reference types="react-scripts" />
+
+export interface Post {
+  userId: number,
+  id: number,
+  title: string,
+  body: string,
+  user: User | null,
+  comments: Comment[] | [],
```
Anna-Tkachenko, line 9: same here, empty array here is redundant

Anna-Tkachenko (changes requested):

src/react-app-env.d.ts
igor-1987: Understood, thank you!

Anna-Tkachenko (approved):
Good job!

PR #1298, opened 2022-10-20
Solution

AniaHna:
(https://AniaHna.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,36 @@
-import React from 'react';
-
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+import { User } from './types/user';
+import { Post } from './types/post';
+import { Comment } from './types/comment';
+import { PostList } from './components/PostList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComments(postId: number): Comment[] {
+  const foundComments = commentsFromServer
+    .filter(comment => comment.postId === postId);
+
+  return foundComments;
+}
+
+export const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: getUser(post.userId),
+  comments: getComments(post.id),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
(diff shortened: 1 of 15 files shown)


ArtemHulkoDev (approved):
everything is fine)
My only suggestion is that there is no need to use ul and li tags when you render array items through the map method, you can see li dots when you open this task in browser

src/App.tsx
```diff
@@ -1,109 +1,36 @@
...
+function getComments(postId: number): Comment[] {
+  const foundComments = commentsFromServer
+    .filter(comment => comment.postId === postId);
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+  return foundComments;
```
Anna-Tkachenko, line 22: you can return the result of filter method without creating additional variable

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,19 @@
...
+import { Comment } from '../../types/comment';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <ul className="CommentList">
+    {comments.map(comment => (
+      <li>
```
Anna-Tkachenko, line 14: key attribute should be on the parent element here (move it to the li element instead of CommentInfo)

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,44 @@
...
+    body,
+    user,
+    comments,
+  },
+}) => (
+  <div className="PostInfo">
+    <div className="PostInfo__header">
+      <h3 className="PostInfo__title">{title}</h3>
+
+      <p>
```
Anna-Tkachenko, line 25: let's not show this paragraph if there is no user

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,19 @@
...
+import { Post } from '../../types/post';
+
+type Props = {
+  posts: Post[];
+};
+
+export const PostList: React.FC<Props> = ({ posts }) => (
+  <ul className="PostList">
+    {posts.map(post => (
+      <li>
```
Anna-Tkachenko, line 14: same here about key attribute

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1318, opened 2022-11-02
solution

mykola8426:


src/App.tsx
```diff
@@ -2,108 +2,33 @@ import React from 'react';
...
+import { PostList } from './components/PostList';
+
+const findUser = (userId: number):User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+};
+
+const findComments = (postId: number):Comment[] => {
+  return [...commentsFromServer].filter(comment => comment.postId === postId);
```
ArtemTopolskyi, line 20: method filter creates new array, so you don't need to create it manually

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,33 @@
...
+    </div>
+
+    <p className="PostInfo__body">
+      {post.body}
+    </p>
+
+    <hr />
+
+    {
+      post.comments.length === 0
```
ArtemTopolskyi, line 28:
```suggestion
      !post.comments.length
```

ArtemTopolskyi (approved):

PR #1324, opened 2022-11-10
Solution

anastasia-babko:
[DEMO LINK](https://anastasia-babko.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,36 @@
 import React from 'react';
-
 import './App.scss';
+import { PostList } from './components/PostList';
+import { Post } from './types/Post';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+function getUser(userId: number): User | null {
+  const newUser = usersFromServer.find(user => user.id === userId);
+
+  return newUser || null;
+}
+
+function getComments(id: number): Comment[] {
+  const newComment = commentsFromServer.filter(comment => (
+    comment.postId === id));
+
+  return newComment || null;
+}
+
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,109 +1,36 @@
...
+}
 
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+function getComments(id: number): Comment[] {
+  const newComment = commentsFromServer.filter(comment => (
+    comment.postId === id));
```
stanislav-topikha, line 19:
```suggestion
  const newComment = commentsFromServer.filter(comment => (
    comment.postId === id
  ));
```

stanislav-topikha (changes requested):
few comments 

remove those dots 
image

src/App.tsx
anastasia-babko: Done.

src/App.tsx
```diff
@@ -1,109 +1,37 @@
...
+
+  return newUser || null;
+}
+
+function getComments(id: number): Comment[] {
+  const newComment = commentsFromServer.filter(comment => (
+    comment.postId === id
+  ));
+
+  return newComment || null;
```
SanyaBratashchuk, line 22:
```suggestion
  return newComment || null;
```
`filter` always return an Array array is a truthy value so getComments never return null

src/types/Post.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from './User';
+import { Comment } from './Comment';
+
+export interface Post {
+  userId: number;
+  id: number;
+  title: string;
+  body: string;
+  user: User | null;
+  comments: Comment[] | [];
```
SanyaBratashchuk, line 10:
```suggestion
  comments: Comment[];
```

SanyaBratashchuk (approved):

src/App.tsx
anastasia-babko: Indeed. Got it, thanks.

PR #1345, opened 2022-11-17
solution

MaxFedak:
[DEMO LINK](https://maxfedak.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,5 +1,45 @@
-import React from 'react';
+import { UserInfo } from '../UserInfo';
+import commentsFromServer from '../../api/comments';
+import usersFromServer from '../../api/users';
+import { Post } from '../../types/post';
+import { CommentList } from '../CommentList';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
-);
+type Props = {
+  post: Post,
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => {
+  const {
+    userId,
+    title,
+    body,
+    id,
+  } = post;
+
+  const user = usersFromServer.find(curUser => curUser.id === userId);
+  const comments = commentsFromServer
+    .filter(curComment => curComment.postId === id);
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{title}</h3>
+
+        <p>
+          {' Posted by  '}
+
+          {user && <UserInfo dataType="post" user={user} />}
+        </p>
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+      {comments.length === 0
+        ? <b data-cy="NoCommentsMessage">No comments yet</b>
+        : <CommentList comments={comments} />}
+    </div>
+  );
+};
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,5 +1,24 @@
-import React from 'react';
+import { User } from '../../types/user';
+import { Comment } from '../../types/comment';
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
-);
+type Props = {
+  user: User | Comment,
+  dataType: 'post' | 'comment',
+};
+
+export const UserInfo: React.FC<Props> = ({ user, dataType }) => {
+  const { email, name } = user;
+  const propperText = dataType === 'comment'
+    ? email
+    : name;
+
+  const propperClass = dataType === 'comment'
+    ? 'CommentInfo__email'
+    : 'UserInfo';
+
+  return (
+    <a className={propperClass} href={`mailto:${email}`}>
+      {propperText}
+    </a>
+  );
+};
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,11 @@
 import React from 'react';
-
 import './App.scss';
-
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import { PostList } from './components/PostList';
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
-
-    <div className="PostList">
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">qui est esse</h3>
-
-          <p>
-            {' Posted by  '}
-
-            <a className="UserInfo" href="mailto:Sincere@april.biz">
-              Leanne Graham
-            </a>
-          </p>
-        </div>
-
-        <p className="PostInfo__body">
-          est rerum tempore vitae sequi sint nihil reprehenderit dolor beatae
-          ea dolores neque fugiat blanditiis voluptate porro vel nihil
-          molestiae ut reiciendis qui aperiam non debitis possimus qui neque
-          nisi nulla
-        </p>
-
-        <hr />
-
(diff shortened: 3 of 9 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,45 @@
...
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+      {comments.length === 0
+        ? <b data-cy="NoCommentsMessage">No comments yet</b>
+        : <CommentList comments={comments} />}
+    </div>
```
masha-girya, line 43: tag `hr` has to be only when there is no comments added, maybe that is why some tests had failed

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,45 @@
...
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{title}</h3>
+
+        <p>
+          {' Posted by  '}
+
+          {user && <UserInfo dataType="post" user={user} />}
```
masha-girya, line 31: interesting, why did you need dataType here? it wasn't in original file

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,24 @@
...
+import { User } from '../../types/user';
+import { Comment } from '../../types/comment';
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
-);
+type Props = {
+  user: User | Comment,
+  dataType: 'post' | 'comment',
+};
```
masha-girya, line 7: maybe here is the problem. You have only one User or null for Post and array of Comments or null per Post. So if you will tie Comments and User to Post from the start, you will have interface of Post, that already includes appropriate User and Comments

src/components/PostInfo/PostInfo.tsx
masha-girya: now I saw, okay)

masha-girya (changes requested):
i believe you can try to change Post interface in the App file, like we did in "Todo List" so it will has a chance to pass tests)

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,45 @@
...
+    userId,
+    title,
+    body,
+    id,
+  } = post;
+
+  const user = usersFromServer.find(curUser => curUser.id === userId);
+  const comments = commentsFromServer
+    .filter(curComment => curComment.postId === id);
+
```
taraselnik, line 22: test for comment you cant to pass because they didn't give in test lucky values ;)
1. they gave on this block new an object wich contains 'user' info and all 'comments'
2. after, you take this values right from .api array  not from parametrs and dont give them chance to get their results which they gave to parametrs. 
the Id they gave is 11 so
you code result gives 5 results
image


they wanted to see 3 results. on post id 11.  (throughout commentList ==>> maping commentInfo == >>>)
image

src/components/PostInfo/PostInfo.tsx
taraselnik: excuse me) i have to ask about i missed it maybe or cant understand what you about ((((

src/components/PostInfo/PostInfo.tsx
taraselnik: nope) this position of hr is perfect

src/components/UserInfo/UserInfo.tsx
taraselnik: commonly there is a sense about passing datas through "the forest" but not in this block.

taraselnik (approved):
Dear Mate, I'd like to put approve cause IMHO you have incredibly structural and nice logic in  your solution which rids a lot of passing datas and don't break brain.

src/components/UserInfo/UserInfo.tsx
taraselnik: > maybe here is the problem. You have only one User or null for Post and array of Comments or null per Post. So if you will tie Comments and User to Post from the start, you will have interface of Post, that already includes appropriate User and Comments

commonly there is a sense about passing datas through "the forest" but not in this block.

src/components/PostInfo/PostInfo.tsx
taraselnik: > interesting, why did you need dataType here? it wasn't in original file

excuse me) i have to ask about i missed it maybe or cant understand what you about ((((

src/components/PostInfo/PostInfo.tsx
taraselnik: > tag `hr` has to be only when there is no comments added, maybe that is why some tests had failed

nope) this position of hr is perfect

taraselnik:
> i believe you can try to change Post interface in the App file, like we did in "Todo List" so it will has a chance to pass tests)

or cheat it)))))

src/components/PostInfo/PostInfo.tsx
MaxFedak: tried both variations, problem is not here)

src/components/PostInfo/PostInfo.tsx
MaxFedak: we use UserInfo each post 2 times for each post with comments
once for author of the post (we show name)
and once for each comment (we show the email)

So i maid a component, which can do both. dataType is used to show which type we need in each iteration

src/components/UserInfo/UserInfo.tsx
MaxFedak: but why problem is here? if u check html structure of the code, it's totally ok
and in the api post has rly 5 comments, i cant. understand why test expects 3

taraselnik: its not here)

src/components/PostInfo/PostInfo.tsx
taraselnik: thanks, I'll try ro reread

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,45 @@
...
+export const PostInfo: React.FC<Props> = ({ post }) => {
+  const {
+    userId,
+    title,
+    body,
+    id,
+  } = post;
+
+  const user = usersFromServer.find(curUser => curUser.id === userId);
+  const comments = commentsFromServer
```
Olexandr6, line 20: You do this logic every time when `PostInfo` rerenders. It's not okay from the performance side. Would be good to prepare the array once in `App.tsx` component and pass it to `<PostList />`

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,45 @@
...
+    .filter(curComment => curComment.postId === id);
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{title}</h3>
+
+        <p>
+          {' Posted by  '}
```
Olexandr6, line 29:
```suggestion
          Posted by
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,24 @@
...
+
+export const UserInfo: React.FC<Props> = ({ user, dataType }) => {
```
Olexandr6, line 9: User info and post info - different things. Would be good to separate your logic to separate components.
In practice, better to split it into small components rather than have one but with complicated logic.

If you need more explanations - feel free to ping me in the slack and discuss this point

Olexandr6 (changes requested):
Almost done👍 Now, let's work on improvements:
Pay attention to styles in `App.scss`, they should be moved to separate files, as well.

Olexandr6 (approved):

PR #1385, opened 2022-12-01
Added the task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,39 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+import { Post } from './types/Post';
+import { Comment } from './types/Comment';
+import { User } from './types/User';
+
+import { PostList } from './components/PostList';
+
+function getComments(postId: number): Comment[] | null {
+  const postComments
+    = commentsFromServer.filter(comment => comment.postId === postId);
+
+  return postComments || null;
+}
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+export const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  comments: getComments(post.id),
+  user: getUser(post.userId),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
 
-    <div className="PostList">
-      <div className="PostInfo">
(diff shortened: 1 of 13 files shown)


src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,21 @@
...
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <div className="CommentList">
+    {comments.map(comment => (
+      <CommentInfo
+        comment={comment}
+        key={comment.id}
+      />
+    ))}
```
Bohdan-Maksymiuk, line 19: use semantic `<ul>` and `<li>` when missing array

src/App.tsx
```diff
@@ -2,108 +2,39 @@ import React from 'react';
...
-        <b data-cy="NoCommentsMessage">No comments yet</b>
-      </div>
+import { PostList } from './components/PostList';
 
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+function getComments(postId: number): Comment[] | null {
```
Bohdan-Maksymiuk, line 15:
```suggestion
function getComments(postId: number): Comment[] {
```
it should work like this
why do you need additional problems with types? In such case there will be just empty array

src/App.tsx
```diff
@@ -2,108 +2,39 @@ import React from 'react';
...
+  const postComments
+    = commentsFromServer.filter(comment => comment.postId === postId);
 
-          <p>
-            {' Posted by  '}
+  return postComments || null;
```
Bohdan-Maksymiuk, line 19:
```suggestion
  return postComments;
```
useless - empty array is always true

Bohdan-Maksymiuk (changes requested):

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,21 @@
...
+  comments: Comment[],
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <ul className="CommentList">
+    {comments.map(comment => (
+      <CommentInfo
+        comment={comment}
+        key={comment.id}
+      />
```
Anastasiia-Svintsova, line 18: would be better to add `<li>` tag in case you use `<ul>` here

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,47 @@
...
+export const PostInfo: React.FC<Props> = ({ post }) => {
+  const {
+    user,
+    title,
+    body,
+    comments,
+  } = post;
+
+  return (
+    <>
```
Anastasiia-Svintsova, line 23: You don't really need this additional wrapper here

Anastasiia-Svintsova (approved):

PR #1400, opened 2022-12-17
react_static-list-of-posts

ysrckr:
DEMO LINK](https://ysr_ckr.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/types/User.ts
+++ b/src/types/User.ts
@@ -1,0 +1,6 @@
+export interface User {
+  id?: number;
+  name: string;
+  username?: string;
+  email: string;
+}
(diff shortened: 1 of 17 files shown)


src/types/User.ts
```diff
@@ -0,0 +1,6 @@
+export interface User {
+  id?: number;
```
UmizDemud, line 2: Data coming from server should have some id in it. I'm not used to any id being omitted. Definitelly not asking you to rewrite here, just wanted to note

UmizDemud (approved):
👍

PR #1416, opened 2022-12-26
Initial commit

artem-hirzhev:
[DEMO LINK](https://artem-hirzhev.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,40 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+import { Post } from './types/Post';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+
+import { PostList } from './components/PostList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComment(id: number): Comment[] {
+  const foundComment = commentsFromServer.filter(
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -2,108 +2,40 @@ import React from 'react';
...
+  const foundComment = commentsFromServer.filter(
+    comment => (id === comment.postId),
+  );
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+  return foundComment || null;
```
SanyaBratashchuk, line 26: filter always return an array so null will never be returned

SanyaBratashchuk: use find instead

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,35 @@
...
+type Props = {
+  comment: Comment;
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => {
+  const {
+    email,
+    body,
+    name,
+  } = comment as Comment;
```
SanyaBratashchuk, line 13:
```suggestion
  } = comment;
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,23 @@
...
-);
+type Props = {
+  user: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  const {
+    name,
+    email,
+  } = user as User;
```
SanyaBratashchuk, line 12: Do not make a destruction if user can be null

SanyaBratashchuk (changes requested):

DarkMistyRoom (approved):

PR #1441, opened 2022-12-28
Complete

SalehAkaJim:
[DEMO LINK](https://SalehBUD.github.io/react_static-list-of-posts/)


src/App.tsx
```diff
@@ -1,109 +1,23 @@
...
+import { PostList } from './components/PostList';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+const posts: Post[] | [] = postsFromServer.map(post => ({
```
Mohameddahi, line 9: The type of posts should only be an array of posts .

src/App.tsx
```diff
@@ -1,109 +1,23 @@
...
+  ...post,
+  user: usersFromServer.find(user => user.id === post.userId) || null,
+  comments: commentsFromServer.filter(comment => comment.postId === post.id),
```
Mohameddahi, line 12: It would be great to create helper functions for being more readable and clear.

dahimohamed (changes requested):
In general , you did great !:)
Please consider those comments.

src/App.tsx
SalehBUD: What if the result of Api is null or error? It's a good practice to have an empty array in order to handle the situation...

dahimohamed (approved):
well done !

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,49 @@
...
+        )}
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+
+      {!comments.length
```
naziamoff, line 38: better to start with positive. Negative booleans are more complicated to understand

naziamoff (approved):
Nice! Just one suggestion

PR #1447, opened 2022-12-28
solution

kenvastolord:
[DEMO LINK](https://kenvastolord.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,34 @@
 import React from 'react';
-
 import './App.scss';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { Post } from './types/Post';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { PostList } from './components/PostList';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getCommentsById(id: number): Comment[] | null {
+  const comments = commentsFromServer.filter(comment => comment.postId === id);
+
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,109 +1,34 @@
...
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+  return foundUser || null;
+}
 
-          <p>
-            {' Posted by  '}
+function getCommentsById(id: number): Comment[] | null {
+  const comments = commentsFromServer.filter(comment => comment.postId === id);
```
Mohameddahi, line 19: Array filter method always will return an array, so there is no need for that returning type "null".

src/types/Post.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from './User';
+import { Comment } from './Comment';
+
+export interface Post {
+  id: number,
+  userId: number,
+  title: string,
+  body: string,
+  comments: Comment[] | null,
```
Mohameddahi, line 9: same here, you can delete that null type.

Mohameddahi (changes requested):
in general, you did well. Just apply those comments.

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,43 @@
...
+          {user && <UserInfo user={user} />}
+        </p>
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      {
+        comments?.length
```
Mohameddahi, line 37: after you removed type null, there is no need for "?"

src/types/Post.ts
Mohameddahi: @kenvastolord  and what about this null?

Mohameddahi (approved):
well done ! consider those comments.

meforoma (approved):
well done @kenvastolord ✅ 

and nice thorough review @Mohameddahi 👍🏼 👍🏼

PR #1461, opened 2022-12-29
React_Static-List-Of-Posts was created

illia-kots:
[DEMO LINK](https://illia-kots.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,35 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+import { Post } from './types/Post';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+
+import { PostList } from './components/PostList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComment(postId: number): Comment[] {
+  return commentsFromServer.filter(comment => comment.postId === postId);
+}
+
+export const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: getUser(post.userId),
+  comments: getComment(post.id),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
-
-    <div className="PostList">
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -2,108 +2,35 @@ import React from 'react';
...
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+import { PostList } from './components/PostList';
 
-          <p>
-            {' Posted by  '}
+function getUser(userId: number): User | null {
```
polosanya, line 15:
```suggestion
function getUserById(userId: number): User | null {
```
image

src/App.tsx
```diff
@@ -2,108 +2,35 @@ import React from 'react';
...
-        </div>
+  return foundUser || null;
+}
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+function getComment(postId: number): Comment[] {
```
polosanya, line 21: Same here

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,19 @@
...
+type Props = {
+  posts: Post[];
+};
+
+export const PostList: React.FC<Props> = ({ posts }) => (
+  <div className="PostList">
+    {posts.map(post => (
+      <PostInfo
+        post={post}
+        key={post.userId}
```
polosanya, line 15: Why do you use userId as a key? 
A few posts can be created by the same user, so they would have the same userId.
I think it's better to use post.id as a key as it's unique for each post.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,24 @@
...
+
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  const {
+    email,
+    name,
+  } = user;
```
polosanya, line 14:
```suggestion
  const { email, name } = user;
```
You can leave it in one line, if there are < 3 props

polosanya (changes requested):
Almost perfect! Please check the comments to improve your code a bit)

src/components/PostList/PostList.tsx
illia-kots: Yea, I'm agree! Ididn't teake this moment into account!Thank's)

SanyaBratashchuk (approved):

PR #1467, opened 2022-12-30
react static list of pists

Mohameddahi:
[DEMO LINK](https://mohameddahi.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,36 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { User } from './Types/users';
+import { Post } from './Types/posts';
+import { Comment } from './Types/comments';
+import { PostList } from './components/PostList';
+
+const getUser = (userId: number): User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+};
+
+const getComment = (postId: number): Comment[] | null => {
+  const foundCommment = commentsFromServer
+    .filter(comment => comment.postId === postId);
+
+  return foundCommment || null;
+};
+
+export const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  comments: getComment(post.id),
+  user: getUser(post.userId),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
-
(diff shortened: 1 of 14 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,48 @@
...
+        </p>
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+
+      {(comments && comments.length > 0)
```
nurdanokcu, line 37:
```suggestion
      {comments.length > 0
```
Filter method is used to find the comments , if there are no comments it will be an empty array so it will always exist , checking the length is enough. .

nurdanokcu (changes requested):
Great job. 
Please see few comments.

src/App.tsx
```diff
@@ -2,108 +2,36 @@ import React from 'react';
...
-            {' Posted by  '}
+  return foundUser || null;
+};
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+const getComment = (postId: number): Comment[] | null => {
```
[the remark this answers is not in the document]
Mohameddahi: Absolutely right, thanks for your review @nurdanokcu

nurdanokcu (approved):
:white_check_mark:

src/App.tsx
```diff
@@ -2,108 +2,36 @@ import React from 'react';
...
-            {' Posted by  '}
+  return foundUser || null;
+};
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+const getComment = (postId: number): Comment[] => {
```
naziamoff, line 19: since the function returns an array - probably this would be more logical to call it in plural?
```suggestion
const getComments = (postId: number): Comment[] => {
```

src/App.tsx
```diff
@@ -2,108 +2,36 @@ import React from 'react';
...
+  const foundCommment = commentsFromServer
```
naziamoff, line 20: same here

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,48 @@
...
+        </p>
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+
+      {comments.length > 0
```
naziamoff, line 37: it will have the same boolean value this way
```suggestion
      {comments.length
```

naziamoff (approved):
Looks good 👍 
Just a couple of suggestions

PR #1493, opened 2023-01-16
Develop

katya-suvorova:
https://katya-suvorova.github.io/react_static-list-of-posts/

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,5 +1,13 @@
 import React from 'react';
+import './UserInfo.scss';
+import { User } from '../../types/User';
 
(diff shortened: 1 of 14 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,20 @@
...
-);
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  const {
+    email,
+    name,
+  } = user;
```
polosanya, line 13:
```suggestion
  const { email, name } = user;
```
You can leave it in one line, if there are < 3 props and the line is not too long

polosanya (approved):
Well done!

PR #1499, opened 2023-01-20
solution

igor-kashchenko:
[DEMO LINK](https://igor-kashchenko.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,11 +16,11 @@ function getUser(userId: number): User | null {
   return foundUser || null;
 }
 
-function getComments(id: number): Comment[] | [] {
+function getComments(id: number): Comment[] {
   const foundComments = commentsFromServer
     .filter(comment => id === comment.postId);
 
-  return foundComments || [];
+  return foundComments;
 }
 
 export const posts: Post[] = postsFromServer.map(post => ({
@@ -29,11 +29,9 @@ export const posts: Post[] = postsFromServer.map(post => ({
   comments: getComments(post.id),
 }));
 
-export const App: React.FC = () => {
-  return (
-    <section className="App">
-      <h1 className="App__title">Static list of posts</h1>
-      <PostList posts={posts} />
-    </section>
-  );
-};
+export const App: React.FC = () => (
+  <section className="App">
+    <h1 className="App__title">Static list of posts</h1>
+    <PostList posts={posts} />
+  </section>
+);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,38 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { Post } from './types/Post';
+import { PostList } from './components/PostList';
 
-export const App: React.FC = () => (
-  <section className="App">
-    <h1 className="App__title">Static list of posts</h1>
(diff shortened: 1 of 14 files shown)


src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,19 @@
...
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  return (
+    <div className="CommentList">
+      {comments.map(comment => (
+        <CommentInfo comment={comment} />
```
Malva37, line 15: you forgot to add key here, its important

Malva37 (approved):
great , but you have to fix s a little.

igor-kashchenko:
> great , but you have to fix s a little.

fixed, ty for your comments!

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+import { Post } from './types/Post';
+import { PostList } from './components/PostList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComments(id: number): Comment[] | [] {
```
Denchepornyuk, line 19: you don't need `| []` here, typescript will understand that function can return Comment[] or empty array

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+  const foundComments = commentsFromServer
+    .filter(comment => id === comment.postId);
+
+  return foundComments || [];
```
Denchepornyuk, line 23: in this case `||` operator will never work, explain to me why 🙂

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+}
+
+export const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: getUser(post.userId),
+  comments: getComments(post.id),
+}));
+
+export const App: React.FC = () => {
+  return (
```
Denchepornyuk, line 33: you can avoid return here

src/components/CommentList/CommentList.scss
```diff
@@ -1 +1,7 @@
-//
+.CommentList {
```
Denchepornyuk, line 1: don't name class with upper case

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,19 @@
...
-  <>Put the list here</>
-);
+import './CommentList.scss';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  return (
```
Denchepornyuk, line 12: avoid return if you can

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+        </p>
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+
+      { comments.length
```
Denchepornyuk, line 37: It's better to check with `> 0`
In some cases with check `comments.length` you can get 0 on the page

denys-cheporniuk (changes requested):

src/App.tsx
igor-kashchenko: > in this case `||` operator will never work, explain to me why 🙂

Operator 'OR' finds first truthy value. However, we assigned to `const foundComments` result of filter action, which can be empty or non-empty array. But because we assigned it to variable, it will always be true (`Boolean(foundComments === []) => true`). Thus, OR operator always returns `foundComments`:)

Denchepornyuk: right

denys-cheporniuk (approved):

PR #1512, opened 2023-01-23
Solution

andrieieva-yuliia:


NVyrsta (approved):
Good job :)

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">
+          {title}
+        </h3>
+
+        <p>
+          {' Posted by  '}
```
alexandra-protyanova, line 27: If user will be `null`, user'll see `Posted by` and that is all 🙂

alexandra-protyanova: Do not render Postred by if user is null

alexandra-protyanova (changes requested):

alexandra-protyanova:
Don't forget to add demo link

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,46 @@
...
+
+          {user === null
```
alexandra-protyanova, line 29:
```suggestion
          {!user
```

alexandra-protyanova (approved):

PR #1513, opened 2023-01-24
Develop

tYuriy:
[DEMO LINK](https://tYuriy.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,16 +12,11 @@ import commentsFromServer from './api/comments';
 import usersFromServer from './api/users';
 
 function getComments(postId: number): Comment[] {
-  const foundComments
-    = commentsFromServer.filter(comment => comment.postId === postId);
-
-  return foundComments;
+  return commentsFromServer.filter(comment => comment.postId === postId);
 }
 
 function getUser(userId: number): User | null {
-  const foundUser = usersFromServer.find(user => user.id === userId);
-
-  return foundUser || null;
+  return usersFromServer.find(user => user.id === userId) || null;
 }
 
 export const totalPostInfo: TotalPostInfo[]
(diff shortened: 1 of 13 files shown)


src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,16 @@
...
-  <>Put the list here</>
+import { TotalPostInfo } from '../../types/TotalPostInfo';
+import { PostInfo } from '../PostInfo';
+
+type Props = {
+  posts: TotalPostInfo[];
+};
+
+export const PostList: React.FC<Props> = ({ posts }) => (
+  <ul className="PostList">
```
TarasHoliuk, line 11: remove default list styles - marks on the left side (for comments and posts):
image

src/App.tsx
```diff
@@ -1,109 +1,39 @@
...
+import usersFromServer from './api/users';
 
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+function getComments(postId: number): Comment[] {
+  const foundComments
+    = commentsFromServer.filter(comment => comment.postId === postId);
```
TarasHoliuk, line 16:
```suggestion
  return commentsFromServer.filter(comment => comment.postId === postId);
```
you can return it immediately instead of creating a new variable

TarasHoliuk: the same with User below

src/types/TotalPostInfo.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from './User';
+import { Comment } from './Comment';
+
+export interface TotalPostInfo {
```
TarasHoliuk, line 4: you can extend `Post` interface instead of duplicating properties

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:+1:

PR #1515, opened 2023-01-24
List of Posts rendered

chermak-oleh:
[DEMO LINK](https://chermak-oleh.github.io/react_static-list-of-posts/)


src/App.tsx
```diff
@@ -2,108 +2,37 @@ import React from 'react';
...
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+  return foundUser || null;
+}
 
-          <p>
-            {' Posted by  '}
+function getCommentsByPostId(PostId: number): Comment[] | [] {
+  const foundComments = commentsFromServer.filter(comment => (
```
NVyrsta, line 20:
```suggestion
  return commentsFromServer.filter(comment => (
```

src/App.tsx
```diff
@@ -2,108 +2,37 @@ import React from 'react';
...
+    comment.postId === PostId));
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+  return foundComments || [];
```
NVyrsta, line 23:
```suggestion

```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,44 @@
...
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+
+      {comments.length > 0
+        ? <CommentList comments={comments} />
+        : (<b data-cy="NoCommentsMessage">No comments yet</b>) }
```
NVyrsta, line 41:
```suggestion
      {comments.length
        ? <CommentList comments={comments} />
        : <b data-cy="NoCommentsMessage">No comments yet</b> }
```

src/types/user.ts
```diff
@@ -0,0 +1,6 @@
+export interface User {
+  id?: number | undefined,
+  name?: string | undefined,
+  username?: string | undefined,
+  email?: string | undefined,
+}
```
NVyrsta, line 6: I don't think that it is a good idea to declare that all parameters are optional
+ we often get data from the format JSON and undefined is not a valid value for JSON

src/types/post.ts
```diff
@@ -0,0 +1,11 @@
+import { Comment } from './Comment';
+import { User } from './User';
+
+export interface Post {
+  userId: number,
+  id: number,
+  title: string,
+  body: string,
+  user?: User | null,
```
NVyrsta, line 9:
```suggestion
  user: User | null,
```

NVyrsta (changes requested):
Nice work 👌
but try to fix problems with tests

src/App.tsx
```diff
@@ -2,108 +2,35 @@ import React from 'react';
...
+import usersFromServer from './api/users';
+import { PostList } from './components/PostList';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { Post } from './types/Post';
+
+function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
```
ArturErmolenko, line 16:
```suggestion
  return usersFromServer.find(user => user.id === userId) || null;
```

ArturErmolenko (approved):
GJ 👍 
Fix formatting

PR #1518, opened 2023-01-24
List of posts first commit

Sergey-Mironenko:


Sergey-Mironenko:
https://Sergey-Mironenko.github.io/react_static-list-of-posts/

Sergey-Mironenko:
Наче все працює як треба, але компонент PostInfo один тест( а саме: should render no comments message is there are no comments) чомусь не проходить

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,48 @@
...
+            </p>
+          </div>
+
+          <p className="PostInfo__body">
+            {body}
+          </p>
+
+          {comments && (
+            <CommentList comments={comments} />
+          )}
```
TarasHoliuk, line 43: to fix tests: check if the `comments.length > 0`. If yes - render list otherwise render `no comments`:
image

src/App.tsx
```diff
@@ -2,108 +2,43 @@ import React from 'react';
...
-        <hr />
+import { Comment } from './types/Comment';
+import { User } from './types/User';
+import { Post } from './types/Post';
+import { PostList } from './components/PostList';
 
-        <b data-cy="NoCommentsMessage">No comments yet</b>
-      </div>
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
```
TarasHoliuk, line 15:
```suggestion
  return usersFromServer.find(user => user.id === userId) || null;
```
no need to create variable just to return it

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,17 @@
...
-  <>Put the list here</>
+import { Post } from '../../types/Post';
+import { PostInfo } from '../PostInfo/PostInfo';
+
+type Props = {
+  posts: Post[];
+};
+
+export const PostList: React.FC<Props> = ({ posts }) => (
+  <div className="PostList">
```
TarasHoliuk, line 11: it's better to use semantic tags for list, i.e `ul` (the same for items)
but don't forget to fix styles

TarasHoliuk (changes requested):

vlad-demchuk (approved):

PR #1519, opened 2023-01-25
Task done. Have a good time!)

Bogdan-Kutasevich:
[DEMO LINK](https://Bogdan-Kutasevich.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,38 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import { User } from './types/User';
+import { Post } from './types/Post';
+import { Comment } from './types/Comment';
+
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { PostList } from './components/PostList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComments(postId: number): Comment[] {
+  const foundComments = commentsFromServer.filter(comment => (
+    comment.postId === postId
+  ));
+
+  return foundComments;
+}
+
+const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: getUser(post.userId),
+  comments: getComments(post.id),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
-
-    <div className="PostList">
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">qui est esse</h3>
-
-          <p>
-            {' Posted by  '}
-
-            <a className="UserInfo" href="mailto:Sincere@april.biz">
-              Leanne Graham
-            </a>
-          </p>
-        </div>
-
-        <p className="PostInfo__body">
-          est rerum tempore vitae sequi sint nihil reprehenderit dolor beatae
-          ea dolores neque fugiat blanditiis voluptate porro vel nihil
-          molestiae ut reiciendis qui aperiam non debitis possimus qui neque
-          nisi nulla
-        </p>
-
-        <hr />
-
-        <b data-cy="NoCommentsMessage">No comments yet</b>
-      </div>
-
-      <div className="PostInfo">
-        <div className="PostInfo__header">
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import { User } from './types/User';
+import { Post } from './types/Post';
+import { Comment } from './types/Comment';
```
maasiimka, line 7: do you need this imports here?

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { PostList } from './components/PostList';
 
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+function getUser(userId: number): User | null {
```
maasiimka, line 14: If you are searching some entity by some filter, specify it in function/method name 
```suggestion
function getUserById(userId: number): User | null {
```
```suggestion
function getUser(userId: number): User | null {
```

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-          <p>
-            {' Posted by  '}
+  return foundUser || null;
```
maasiimka, line 17: You don't need `const` here. You can return value immediately.

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+  const foundComments = commentsFromServer.filter(comment => (
+    comment.postId === postId
+  ));
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+  return foundComments;
```
maasiimka, line 25: Same situation as previous

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
-              quia molestiae reprehenderit quasi aspernatur aut expedita
-              occaecati aliquam eveniet laudantium omnis quibusdam delectus
-              saepe quia accusamus maiores nam est cum et ducimus et vero
-              voluptates excepturi deleniti ratione
-            </div>
-          </div>
-        </div>
-      </div>
-    </div>
+export const App: React.FC = () => (
```
maasiimka, line 34:
```suggestion
export const App = () => (
```
```suggestion
export const App: React.FC = () => (
```

maasiimka (approved):
You should split `scss` styles on different files. But I think it's not critical. Good job!

src/App.tsx
denys-cheporniuk: agree `getUserById` is better name for this function

src/App.tsx
denys-cheporniuk: `const` can be here

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+  <section className="App">
+    <h1 className="App__title">Static list of posts</h1>
+    <PostList posts={posts} />
```
denys-cheporniuk, line 37:
```suggestion
    <h1 className="App__title">Static list of posts</h1>

    <PostList posts={posts} />
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,20 @@
 import React from 'react';
+import { User } from '../../types/User';
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
+type Props = {
+  user: User | null;
+};
```
denys-cheporniuk, line 6: for UserInfo component user can't be null

denys-cheporniuk (changes requested):

src/App.tsx
Bogdan-Kutasevich: yep, they describe return function type)

src/App.tsx
Bogdan-Kutasevich: agree too

Bogdan-Kutasevich:
> You should split `scss` styles on different files. But I think it's not critical. Good job!

Thanks for review)

denys-cheporniuk (approved):

PR #1527, opened 2023-01-25
Solution

maasiimka:


src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,31 @@
...
 
-export const CommentInfo: React.FC = () => (
-  <>Put the comment here</>
-);
+type Props = {
+  comment: Comment,
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => {
+  const { name, email, body } = comment;
```
oleh-holovnykh, line 9: improve readability
```suggestion
export const CommentInfo: React.FC<Props> = ({ comment }) => {
  const { 
    name, 
    email, 
    body,
   } = comment;
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,13 @@
...
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
+type Props = {
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user.email}`}>
+    {user.name}
+  </a>
```
oleh-holovnykh, line 12: destructuring everywhere)

```suggestion
 const {
    name,
    email,
  } = user;

  <a className="UserInfo" href={`mailto:${email}`}>
    {name}
  </a>
```

src/types/Post.ts
```diff
@@ -0,0 +1,11 @@
+import { Comment } from './Comment';
+import { User } from './User';
+
+export interface Post {
+  userId: number,
+  id: number,
+  title: string,
+  body: string,
+  user: User | undefined,
+  comments: Comment[] | undefined,
```
oleh-holovnykh, line 10: comments will not be undefined. Filter will return an empty array
```suggestion
  comments: Comment[],
```

oleh-holovnykh (changes requested):
Good job! 👍

src/App.tsx
```diff
@@ -2,108 +2,25 @@ import React from 'react';
...
+  const { userId, id } = post;
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+  return {
+    ...post,
+    user: usersFromServer.find((user) => userId === user.id),
```
Anna-Tkachenko, line 16: let's create separate function that will search the user
and will be better to set null if there is no user on server
because `undefined` is some default value, and it's not clear if it wasn't found on server or your object just doesn't have such property

src/App.tsx
```diff
@@ -2,108 +2,25 @@ import React from 'react';
...
+    comments: commentsFromServer.filter(({ postId }) => postId === id),
```
Anna-Tkachenko, line 17: same here, will be better to create separate function that will return needed comments

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,17 @@
...
-  <>Put the list here</>
+import { Comment } from '../../types/Comment';
+
+type Props = {
+  comments: Comment[]
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <ul className="CommentList">
+    {comments.map((comment) => (
```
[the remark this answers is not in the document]
Anna-Tkachenko: awesome that you used semantic tags, you can add small fixed to the design for disabling these dots

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,43 @@
...
+type Props = {
+  post: Post,
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => {
+  const {
+    title,
+    body,
+    user,
+    comments = [],
```
Anna-Tkachenko, line 17: it's not clear why did you set an empty array here by default
`comments` property in post entity can't be undefined

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,43 @@
...
+  return (
+    <li className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{title}</h3>
+
+        <p>
+          {' Posted by  '}
+
+          {user && <UserInfo user={user} />}
+        </p>
```
Anna-Tkachenko, line 29: will be better don't render all paragraph if there is no user

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
-              </a>
-            </div>
-
-            <div className="CommentInfo__body">
-              dolorum voluptas laboriosam quisquam ab totam beatae et aut
-              aliquid optio assumenda voluptas velit itaque quidem voluptatem
-              tempore cupiditate in itaque sit molestiae minus dolores magni
-            </div>
-          </div>
+const findUserById = (users: User[], id: number): FoundUser => (
```
Anna-Tkachenko, line 16: you don't need to pass users array as an argument, you can take it in this function directly from `usersFromServer` variable

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+  users.find((user) => id === user.id) || null
+);
 
-          <div className="CommentInfo">
-            <div className="CommentInfo__title">
-              <strong className="CommentInfo__name">
-                odio adipisci rerum aut animi
-              </strong>
+const filterCommentsById = (comments: Comment[], id: number) => (
```
Anna-Tkachenko, line 20: will be better to name this function like `getFilteredCommentsBy...`
because in this function you don't just filter the comments, the main purpose - is to get the needed comments for corresponding post

and it's not clear if of what entity you passed as the second argument, will be better to specify it

Anna-Tkachenko (changes requested):

src/App.tsx
maasiimka: So why would I make function? Logic isn't reusable in this case. I could use filter to make new constants in the way the post was created. It would be the same result

Anna-Tkachenko: it's great approach to create separate functions for separate purposes
it's not critical to fix this point here with users, you created good function that can be used in many places, and that will accept different arrays and will find the needed user

just here we have only one general users array, so you also was able just directly search in those array

src/App.tsx
```diff
@@ -2,108 +2,38 @@ import React from 'react';
...
+const findUserById = (users: User[], id: number): FoundUser => (
+  users.find((user) => id === user.id) || null
+);
 
-          <div className="CommentInfo">
-            <div className="CommentInfo__title">
-              <strong className="CommentInfo__name">
-                odio adipisci rerum aut animi
-              </strong>
+const getFilteredCommentsById = (comments: Comment[], id: number) => (
```
Anna-Tkachenko, line 20: it's not clear what id you passed as an argument
will be better to name it `postId`, as you want to find comments for some post

Anna-Tkachenko (approved):

PR #1535, opened 2023-01-25
solution

Kravchenko71213:
https://Kravchenko71213.github.io/react_static-list-of-posts/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,36 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { Post } from './types/Post';
+import { PostList } from './components/PostList';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComments(id: number): Comment[] {
+  const foundComments = commentsFromServer
+    .filter(comment => id === comment.postId);
+
+  return foundComments;
+}
+
+export const posts: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: getUser(post.userId),
+  comments: getComments(post.id),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
-
-    <div className="PostList">
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">qui est esse</h3>
-
-          <p>
-            {' Posted by  '}
-
-            <a className="UserInfo" href="mailto:Sincere@april.biz">
-              Leanne Graham
-            </a>
-          </p>
-        </div>
-
-        <p className="PostInfo__body">
-          est rerum tempore vitae sequi sint nihil reprehenderit dolor beatae
-          ea dolores neque fugiat blanditiis voluptate porro vel nihil
-          molestiae ut reiciendis qui aperiam non debitis possimus qui neque
-          nisi nulla
-        </p>
-
-        <hr />
-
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -2,108 +2,36 @@ import React from 'react';
...
+  return foundUser || null;
+}
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+function getComments(id: number): Comment[] {
+  const foundComments = commentsFromServer
```
NVyrsta, line 20:
```suggestion
  return  commentsFromServer
```

src/App.tsx
```diff
@@ -2,108 +2,36 @@ import React from 'react';
...
+    .filter(comment => id === comment.postId);
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+  return foundComments;
```
NVyrsta, line 23:
```suggestion
   
```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,41 @@
...
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{title}</h3>
+
+        <p>
+          {' Posted by  '}
+          { user && <UserInfo user={user} />}
+        </p>
```
NVyrsta, line 27: if user is null we will see {' Posted by '} ...nothing

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,41 @@
...
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+
+      { comments.length > 0
```
NVyrsta, line 36:
```suggestion
      { comments.length
```

NVyrsta (changes requested):
Good job ╰(*°▽°*)╯
but have a little to fix

src/App.tsx
```diff
@@ -2,108 +2,34 @@ import React from 'react';
...
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { Post } from './types/Post';
+import { PostList } from './components/PostList';
+
+function getUser(userId: number): User | null {
```
Anna-Tkachenko, line 13: will be better to name this function like `findUser`, because you can't be sure that you get some object, you can return null from this function

src/App.tsx
```diff
@@ -2,108 +2,34 @@ import React from 'react';
...
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComments(id: number): Comment[] {
```
Anna-Tkachenko, line 19: it's not clear id of what entity you passed here for searching comments

src/components/PostInfo/PostInfo.tsx
Anna-Tkachenko: agree, fix @NVyrsta comment
you shouldn't render this paragraph if there is no user for this post

Anna-Tkachenko (changes requested):

src/App.tsx
Anna-Tkachenko: but this name was here by default, so it's not critical to fix it

src/App.tsx
```diff
@@ -2,108 +2,34 @@ import React from 'react';
...
+import { Post } from './types/Post';
+import { PostList } from './components/PostList';
+
+function findUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getComments(commentId: number): Comment[] {
```
Anna-Tkachenko, line 19: it's not comment id, check what you passed here

Anna-Tkachenko (approved):
Good job!

PR #1619, opened 2023-03-10
initial solution

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,13 +13,13 @@ import { User } from './types/User';
 import { Comment } from './types/Comment';
 import { PostList } from './components/PostList';
 
-function getUser(id: number): User | null {
+function getUserById(id: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === id);
 
   return foundUser || null;
 }
 
-function getComments(id: number): Comment[] {
+function getCommentsById(id: number): Comment[] {
   return commentsFromServer.filter(comment => (
     comment.postId === id
   ));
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,109 +1,39 @@
...
+
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { Post } from './types/Post';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { PostList } from './components/PostList';
+
+function getUser(id: number): User | null {
```
ivbutsykin, line 16: It's small point in checklist: https://github.com/mate-academy/react_static-list-of-posts/blob/master/checklist.md [6 point]
```suggestion
function getUserById(id: number): User | null {
```

src/App.tsx
```diff
@@ -1,109 +1,39 @@
...
+  const foundUser = usersFromServer.find(user => user.id === id);
+
+  return foundUser || null;
```
ivbutsykin, line 19: I thought about if this function should return `null` or `undefined` by default. And yes, I like your idea to return `null`. `Post` seems better with `user: null` type

ivbutsykin (approved):
Although, you can see a lot of comments here, they are not critical. Make attention on this (especially `...byId` naming and` <span />`).

Good job!

PR #1630, opened 2023-03-13
commit message

diamat1997:


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,22 @@
...
-);
+import { User } from '../../types/User';
+import './UserInfo.scss';
+
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  const { name, email } = user;
```
ivbutsykin, line 11: You have different destructuring  styles. For example, here right in props:
```
export const CommentInfo: React.FC<Props> = ({
  comment: {
    name,
    email,
    body,
  },
})
```

I suggest to choose one. As for me, I prefer first variant: `const { name, email } = user;`

ivbutsykin (approved):
Good job! 
Make attention on comments please)

PR #1631, opened 2023-03-13
static posts list was created

rashyd-hasratov:
[DEMO LINK](https://rashyd-hasratov.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,109 +1,33 @@
 import React from 'react';
+import { User } from './types/User';
+import { PostList } from './components/PostList';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+const getUserById = (userId: number): User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+};
+
+const posts = postsFromServer.map(post => (
+  {
+    ...post,
+    user: getUserById(post.userId),
+    comments: commentsFromServer.filter(comment => (
+      comment.postId === post.id
+    )),
+  }
+));
 
 export const App: React.FC = () => (
   <section className="App">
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,109 +1,33 @@
...
-            </a>
-          </p>
-        </div>
+const posts = postsFromServer.map(post => (
+  {
+    ...post,
+    user: getUserById(post.userId),
+    comments: commentsFromServer.filter(comment => (
+      comment.postId === post.id
+    )),
```
KuharchukDmytro, line 23: better to make helper function for comments filtering

src/App.tsx
```diff
@@ -1,109 +1,33 @@
...
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">
-            doloremque illum aliquid sunt
-          </h3>
+const getUserById = (userId: number): User | null => {
+  const foundUser = usersFromServer.find(user => user.id === userId);
 
-          <p>
-            {' Posted by  '}
+  return foundUser || null;
```
KuharchukDmytro, line 14: on my opinion, there cant be post without user, so if we havent the user we can throw en error, and such way in types we can write just
`
user: User
`

src/App.tsx
```diff
@@ -1,109 +1,33 @@
...
+};
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+const posts = postsFromServer.map(post => (
```
KuharchukDmytro, line 17:
```suggestion
const posts: Post[] = postsFromServer.map(post => (
```

KuharchukDmytro (approved):
there is only one issue with user validation, but it only on my opinion, in task we havent such a punct, so good job👍

src/App.tsx
rashyd-hasratov: Yeah, but I thought that there can be a situation when post was created by person, that is already deleted from database, that's why I added the check

DenysDubas (approved):

sergii-nosachenko (changes requested):
Quick fix

PR #1650, opened 2023-03-14
Solution

VadymTsyndra:
https://VadymTsyndra.github.io/react_static-list-of-posts/

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,5 +1,38 @@
 import React from 'react';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+import './PostInfo.scss';
+
+import commentsFromServer from '../../api/comments';
+import { Post } from '../../types/post';
+import { UserInfo } from '../UserInfo';
+import { CommentList } from '../CommentList';
+
+type Props = {
+  post: Post;
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <div className="PostInfo">
+    <div className="PostInfo__header">
+      <h3 className="PostInfo__title">{post.title}</h3>
+
+      <p>
+        {' Posted by  '}
+
+        <UserInfo user={post.user} />
+      </p>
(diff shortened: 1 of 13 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,38 @@
...
+    </div>
+
+    <p className="PostInfo__body">
+      {post.body}
+    </p>
+
+    <hr />
+
+    {post.comment ? (
+      <CommentList comments={commentsFromServer} />
```
nazarmatsevych, line 33:
```suggestion
      <CommentList comments={post.comment} />
```

Now in your demo you have all posts in one post.

nazarmatsevych (changes requested):
Almost done but every post should be unique, pls check my comment below

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,38 @@
...
+    </div>
+
+    <p className="PostInfo__body">
+      {post.body}
+    </p>
+
+    <hr />
+
+    {post.comment ? (
+      <CommentList posts={posts} />
```
ivbutsykin, line 33: Hm, I think some mistake here, your `<CommentList />` accept `posts` but as I understand it should receive `comment`, no? 

Due to this you can compare posts on your page and see same content in comment blocks.

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,38 @@
...
-  <>Put the post here</>
+import './PostInfo.scss';
+
+import { Post } from '../../types/post';
+import { UserInfo } from '../UserInfo';
+import { CommentList } from '../CommentList';
+
+type Props = {
+  post: Post;
+  posts: Post[];
```
ivbutsykin, line 11: So after rewrite, no need will be in this prop

ivbutsykin (changes requested):

slavanosatskyi (approved):

PR #1653, opened 2023-03-14
Develop

YevheniiHura:
https://YevheniiHura.github.io/react_static-list-of-posts/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,13 +3,12 @@ import { PostList } from './components/PostList';
 
 import './App.scss';
 
-import { Post } from './types/Post';
-import { User } from './types/User';
-import { Comment } from './types/Comment';
-
-import postsFromServer from './api/posts';
-import commentsFromServer from './api/comments';
-import usersFromServer from './api/users';
+import { Post, Comment, User } from './types/index';
+import {
+  postsFromServer,
+  commentsFromServer,
+  usersFromServer,
+} from './api/index';
 
 function getUser(userId: number): User | null {
   const findUser = usersFromServer.find(user => user.id === userId);
@@ -18,11 +17,9 @@ function getUser(userId: number): User | null {
 }
 
 function getComment(postId: number): Comment[] {
-  const findComment = commentsFromServer.filter(
+  return commentsFromServer.filter(
     comment => comment.postId === postId,
   );
-
-  return findComment;
 }
 
 export const posts: Post[] = postsFromServer.map(post => ({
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,109 +1,40 @@
...
+import { PostList } from './components/PostList';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import { Post } from './types/Post';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
```
OlEzhA0, line 8: you can add 
`index.ts` file 

and add exports
```
export * from './Post'
export * from './User'
export * from './Comment'
```

and use import like

```
import { Post, User, Comment } from './types'
```

src/App.tsx
```diff
@@ -1,109 +1,40 @@
...
-          nisi nulla
-        </p>
-
-        <hr />
-
-        <b data-cy="NoCommentsMessage">No comments yet</b>
-      </div>
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
```
OlEzhA0, line 12: the same as above

src/App.tsx
```diff
@@ -1,109 +1,40 @@
...
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+function getComment(postId: number): Comment[] {
+  const findComment = commentsFromServer.filter(
+    comment => comment.postId === postId,
+  );
```
OlEzhA0, line 23:
```suggestion
  return commentsFromServer.filter(
    comment => comment.postId === postId,
  );
```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+          <UserInfo user={user} />
+        )}
+      </p>
+    </div>
+
+    <p className="PostInfo__body">
+      {body}
+    </p>
+
+    {comments.length > 0
```
OlEzhA0, line 38:
```suggestion
    {comments.length
```

OlEzhA0 (changes requested):
hey! I left few recommendations
would be better to apply it

src/App.tsx
YevheniiHura: done

src/components/PostInfo/PostInfo.tsx
YevheniiHura: done

src/App.tsx
YevheniiHura: done

IvanFesenko (approved):

PR #1669, opened 2023-03-16
add task solution

nastychenko:
https://nastychenko.github.io/react_static-list-of-posts/

### DIFF
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,5 +1,45 @@
 import React from 'react';
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
-);
+import './CommentList.scss';
+
+import { CommentInfo } from '../CommentInfo/CommentInfo';
+
+import { Comment } from '../../types/Comment';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  if (comments.length === 0) {
+    return (
+      <>
+        <hr />
+        <b data-cy="NoCommentsMessage">No comments yet</b>
+      </>
+    );
+  }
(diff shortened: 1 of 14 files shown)


src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,36 @@
...
+        comments,
+      } = post;
+
+      return (
+        <li className="PostInfo" key={id}>
+          <PostInfo
+            title={title}
+            body={body}
+            user={user}
+            comments={comments}
```
polosanya, line 30: Just pass the whole post and get all the needed props inside

src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,45 @@
...
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => {
+  if (comments.length === 0) {
+    return (
+      <>
+        <hr />
+        <b data-cy="NoCommentsMessage">No comments yet</b>
+      </>
+    );
```
polosanya, line 20: Make this check in PostInfo

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,43 @@
...
+          {' Posted by  '}
+          <UserInfo user={user} />
+        </p>
+      )}
+    </div>
+
+    <p className="PostInfo__body">
+      {body}
+    </p>
+    <CommentList comments={comments} />
```
polosanya, line 41: You should check if comments are present and return error message if not.
Don't render CommentList with empty comments
Use ternary operator here

polosanya (changes requested):
Let's try to pass all tests
image

nazarmatsevych (approved):
👍

PR #1698, opened 2023-04-06
done

andrudenko:
[DEMO LINK](https://andrudenko.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentList/CommentList.scss
+++ b/src/components/CommentList/CommentList.scss
@@ -1,1 +1,7 @@
-//
+.CommentList {
+  display: flex;
+  flex-direction: column;
+  gap: 0.7em;
+  background-color: #eee;
+  padding: 1em;
+}
--- a/src/components/PostInfo/PostInfo.scss
+++ b/src/components/PostInfo/PostInfo.scss
@@ -1,1 +1,16 @@
-//
+.PostInfo {
+  margin: 10px auto;
+  width: 90%;
+  border: 1px solid #000;
+  border-radius: 8px;
+  background-color: lightgray;
+  padding: 1em;
+
+  &__title {
+    margin: 0;
+  }
+
+  &__header {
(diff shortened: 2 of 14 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,17 @@
...
+import { User } from '../../types/User';
+import './UserInfo.scss';
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
```
etojeDenys, line 9: Use array destructuring to destructure the `user` prop. For example: `export const UserInfo: React.FC<Props> = ({ user: { name, email } }) => (`

src/components/CommentList/CommentList.scss
```diff
@@ -1 +1,7 @@
-//
```
etojeDenys, line 1: Use line comments instead of block comments. Replace '//' with '// CommentList styles' to provide more context.

src/components/PostInfo/PostInfo.scss
```diff
@@ -1 +1,16 @@
-//
+.PostInfo {
+  margin: 10px auto;
```
etojeDenys, line 2: Avoid using inline styles like 'width: 90%;'. Consider creating a reusable CSS class or using CSS variables to handle common styling.

src/components/UserInfo/UserInfo.scss
```diff
@@ -1 +1,3 @@
-//
```
etojeDenys, line 1: Use variables for colors and other repeated values. Consider creating a separate file for variables and importing it in your stylesheets.

etojeDenys (approved):
LGTM

PR #1700, opened 2023-04-06
Solution

yana-kotsulym:
https://yana-kotsulym.github.io/react_static-list-of-posts/

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,5 +1,47 @@
 import React from 'react';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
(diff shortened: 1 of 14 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,47 @@
...
+  const {
+    id,
+    title,
+    body,
+    user,
+    comments,
+  } = post;
+
+  return (
+    <div className="PostInfo" key={id}>
```
SanyaBratashchuk, line 23:
```suggestion
    <div className="PostInfo" >
```
There is no need to add key here

SanyaBratashchuk (approved):

PR #1761, opened 2023-04-17
Solution

kate-matviiv:


polosanya (changes requested):
Please don't forget to add a demo link to the PR description, so we can properly check your work
image

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,40 @@
...
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">
+          {title}
+        </h3>
+
+        <p>
+          {' Posted by  '}
+
+          {user && <UserInfo user={user} />}
+        </p>
```
anstsot, line 29: it will be better to wrap the whole `p` tag and check if user exists

anstsot (approved):
Great job, approved!

PR #1791, opened 2023-05-07
Develop

ctgdprzyb:
[DEMO LINK](https://ctgdprzyb.github.io/react_static-list-of-posts/)

I've technically managed to get the site to render properly, but I've yet to split CSS styles to separate files, and I implemented comments wrong (as in, they work, but it's not done the way the task expects me to - I glossed over that bit).

I'll fix that, but for now here's the working thing.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,9 +2,24 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
+import { Post } from './types/Post';
+import { User } from './types/User';
+
+import postsFromServer from './api/posts';
 // import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import usersFromServer from './api/users';
+
+function getUser(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  // if there is no user with a given userId
+  return foundUser || null;
+}
+
+export const todos: Post[] = postsFromServer.map(post => ({
+  ...post,
+  user: getUser(post.userId),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,6 +8,7 @@ import { User } from './types/User';
 import postsFromServer from './api/posts';
 // import commentsFromServer from './api/comments';
 import usersFromServer from './api/users';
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -2,108 +2,43 @@ import React from 'react';
...
-
-        <hr />
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { PostList } from './components/PostList';
 
-        <b data-cy="NoCommentsMessage">No comments yet</b>
-      </div>
+function getUser(userId: number): User | null {
```
d-wasilewski, line 14: function names, checklist

src/App.tsx
```diff
@@ -2,108 +2,43 @@ import React from 'react';
...
+  ));
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+  if (comments.length > 0) {
+    return comments;
+  }
```
d-wasilewski, line 27: you dont need to do this, you can return an array even when it's empty. This will probably solve a lot of problems here

src/App.tsx
```diff
@@ -2,108 +2,43 @@ import React from 'react';
...
-              quia molestiae reprehenderit quasi aspernatur aut expedita
-              occaecati aliquam eveniet laudantium omnis quibusdam delectus
-              saepe quia accusamus maiores nam est cum et ducimus et vero
-              voluptates excepturi deleniti ratione
-            </div>
-          </div>
-        </div>
-      </div>
-    </div>
+    <PostList postList={posts} />
```
d-wasilewski, line 42:
```suggestion
    <PostList posts={posts} />
```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,44 @@
...
+
+        {user ? <UserInfo userInfo={user} /> : undefined}
+      </p>
+    </div>
+
+    <p className="PostInfo__body">
+      {body}
+    </p>
+
+    { comments
```
d-wasilewski, line 31: check for `comment.length`. It's the only way to check if an array is truthy or falsy in the way you want it

d-wasilewski (changes requested):

PR #1804, opened 2023-05-11
solution

Roman-Kulchytskyi:
https://roman-kulchytskyi.github.io/react_static-list-of-posts/


src/components/UserInfo/UserInfo.scss
```diff
@@ -1 +1,3 @@
-//
```
etojeDenys, line 1: Consider removing the empty comment line, as it serves no purpose.

etojeDenys (changes requested):

PR #1806, opened 2023-05-12
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentInfo/CommentInfo.tsx
+++ b/src/components/CommentInfo/CommentInfo.tsx
@@ -1,5 +1,30 @@
 import React from 'react';
 
-export const CommentInfo: React.FC = () => (
-  <>Put the comment here</>
+import { Comment } from '../../types/Comment';
+
+type Props = {
+  comment: Comment,
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => (
+  <>
+    <div className="CommentInfo__title">
+      <strong className="CommentInfo__name">
+        {comment.name}
+      </strong>
+
+      {' by '}
+
+      <a
+        className="CommentInfo__email"
+        href={`mailto:${comment.email}`}
+      >
+        {comment.email}
+      </a>
+    </div>
+
+    <div className="CommentInfo__body">
+      {comment.body}
+    </div>
+  </>
 );
--- a/src/components/PostList/PostList.tsx
+++ b/src/components/PostList/PostList.tsx
@@ -1,5 +1,44 @@
 import React from 'react';
 
-export const PostList: React.FC = () => (
-  <>Put the list here</>
+import { PostInfo } from '../PostInfo';
+import { Post } from '../../types/Post';
+import { User } from '../../types/User';
+import { Comment } from '../../types/Comment';
+
+type Props = {
+  posts: Post[],
+  comments: Comment[],
+  users: User[],
+};
+
+export const PostList: React.FC<Props> = ({
+  posts,
+  users,
+  comments,
+}) => (
+  <div className="PostList">
+    {posts.map(postItem => {
+      const user = users.find(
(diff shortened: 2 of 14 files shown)


src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,30 @@
...
+
+type Props = {
+  comment: Comment,
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => (
+  <>
+    <div className="CommentInfo__title">
+      <strong className="CommentInfo__name">
+        {comment.name}
```
agniya-i, line 13: Use destructuring for `comment` properties

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,22 @@
...
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <p>
+    {' Posted by  '}
+
+    <a
+      className="UserInfo"
+      href={`mailto:${user.email}`}
+    >
+      {user.name}
```
agniya-i, line 19: Same here

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,44 @@
...
+      if (!user) {
+        return null;
+      }
+
+      const postWithComments = {
+        ...postItem,
+        user,
+        comments: postComments,
+      };
+
```
agniya-i, line 36: It would be better to get all the necessary data outside `map` 
You can create a separate function to prepare an array with posts and use this array for mapping

agniya-i (changes requested):

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,67 @@
...
+import { User } from '../../types/User';
+import { Comment } from '../../types/Comment';
+
+type Props = {
+  posts: Post[],
+  comments: Comment[],
+  users: User[],
+};
+
+function createPostsInfo(
```
agniya-i, line 14: It would be better to create this function in `App` component and pass to the `PostList` only array of posts, instead of `posts, users, comments`

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,67 @@
...
+  posts: Post[],
+  comments:Comment[],
+  users: User[],
+) {
+  const postsInfo = posts.map(post => {
+    const user = users.find(
```
agniya-i, line 20: Create a separate functions to find users and posts to make this `createPostsInfo` function simpler

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,67 @@
...
+
+    if (!user) {
+      return null;
+    }
+
+    return {
+      ...post,
+      user,
+      comments: postComments,
+    };
```
agniya-i, line 35: So you can use these functions here, for example: 

```suggestion
    return {
      ...post,
      user: getUser(post.userId),
      comments: getComments(post.id),
    };
```

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,67 @@
...
+    const user = users.find(
+      (person) => person.id === post.userId,
+    );
+    const postComments = comments.filter(
+      (item) => post.id === item.postId,
+    );
+
+    if (!user) {
+      return null;
+    }
```
agniya-i, line 29: In case there is no user, it would be better to set `null` to user field instead to whole item

agniya-i (changes requested):
Let's improve a little bit!)

agniya-i (approved):
👍

PR #1846, opened 2023-06-07
add solution

ivankhrsh:
[DEMO LINK](https://ivankhrsh.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,108 +2,40 @@ import React from 'react';
 
 import './App.scss';
 
-// import postsFromServer from './api/posts';
-// import commentsFromServer from './api/comments';
-// import usersFromServer from './api/users';
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { PostList } from './components/PostList';
+import { PostPrepared } from './types/postsPrepared';
+
+function getUserById(userId: number): User | null {
+  const foundUser = usersFromServer.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
+function getCommentById(postId: number): Comment[] | null {
+  const foundComments = commentsFromServer
+    .filter(comment => comment.postId === postId);
+
+  return foundComments || null;
+}
+
+export const postsPrepared: PostPrepared[] = postsFromServer.map(post => ({
+  ...post,
+  user: getUserById(post.userId),
+  comments: getCommentById(post.id),
+}));
 
 export const App: React.FC = () => (
   <section className="App">
     <h1 className="App__title">Static list of posts</h1>
 
     <div className="PostList">
-      <div className="PostInfo">
-        <div className="PostInfo__header">
-          <h3 className="PostInfo__title">qui est esse</h3>
-
-          <p>
-            {' Posted by  '}
-
-            <a className="UserInfo" href="mailto:Sincere@april.biz">
-              Leanne Graham
-            </a>
-          </p>
-        </div>
-
-        <p className="PostInfo__body">
-          est rerum tempore vitae sequi sint nihil reprehenderit dolor beatae
-          ea dolores neque fugiat blanditiis voluptate porro vel nihil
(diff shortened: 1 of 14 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{post.title}</h3>
+
+        <p>
+          {' Posted by  '}
+
+          <UserInfo key={post.userId} user={(post.user)} />
+        </p>
```
33Orange, line 26: If text `posted by` rendered only for userinfo maybe would be better to move it inside `UserInfo` component ?

Why are you rendering component user when it doesn't exist ?
Let's add check for `user` here before render it.

```suggestion
          {user && <UserInfo key={post.userId} user={user} />}
```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+      <hr />
+
+      {(post.comments && post.comments?.length > 0)
+        ? <CommentList comments={post.comments} />
+        : <b data-cy="NoCommentsMessage">No comments yet</b>}
+
+    </div>
+
+  );
+};
```
33Orange, line 42: Let's use destructuring assignment here for post.

33Orange (changes requested):
Almost done

A few moments to fix

src/App.tsx
```diff
@@ -2,108 +2,40 @@ import React from 'react';
...
-            {' Posted by  '}
+  return foundUser || null;
+}
 
-            <a className="UserInfo" href="mailto:Julianne.OConner@kory.org">
-              Patricia Lebsack
-            </a>
-          </p>
-        </div>
+function findCommentById(postId: number): Comment[] | null {
```
dinaminahi, line 20:
```suggestion
function findCommentsByPostId(postId: number): Comment[] | null {
```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,41 @@
...
+    body,
+    comments,
+  } = post;
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{title}</h3>
+
+        {user && <UserInfo key={userId} user={(user)} />}
```
dinaminahi, line 26: No need of `key` here. Keys should be given to the elements inside the array to give the elements a stable identity

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,41 @@
...
+      </div>
+
+      <p className="PostInfo__body">
+        {body}
+      </p>
+
+      <hr />
+
+      {(comments && comments?.length > 0)
```
dinaminahi, line 35: You can have only the check for it's length here

src/App.tsx
```diff
@@ -2,108 +2,40 @@ import React from 'react';
...
+  const foundComments = commentsFromServer
+    .filter(comment => comment.postId === postId);
```
dinaminahi, line 22: Can we move these helper functions to a separate file?

dinaminahi (changes requested):
Good job! Few fixes needed

src/components/PostInfo/PostInfo.tsx
ivankhrsh: image
don't get it, is there any other way to fix it?

src/types/postsPrepared.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from './User';
+import { Comment } from './Comment';
+
+export interface PostPrepared {
+  userId: number,
+  id: number,
+  title: string,
+  body: string,
+  user: User | null,
+  comments?: Comment[] | null,
```
dinaminahi, line 10: Let it be empty array if there are no comments, don't set it to null in your filter fn
And if you make it optional it means it can be undefined, but in our case it is either empty array or array of comments 
```suggestion
  comments: Comment[] ,
```

33Orange (approved):
Well done

PR #1866, opened 2023-06-08
solution

snuber-w:
- [DEMO LINK](https://snuber-w.github.io/react_static-list-of-posts/)


silchyn (changes requested):
And again, that's the problem to solve

silchyn (approved):

PR #1894, opened 2023-06-19
solution

Gosyav:
https://vo7kov.github.io/react_static-list-of-posts/

### DIFF
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,5 +1,27 @@
+// Libraries
 import React from 'react';
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
+// Styles
+import './CommentList.scss';
+
+// Types
+import { Comment } from '../../types/Comment';
+
+// Components
+import { CommentInfo } from '../CommentInfo';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <ul className="CommentList">
+    {comments.map((comment: Comment) => (
+      <CommentInfo
+        commentName={comment.name}
+        email={comment.email}
+        commentBody={comment.body}
+      />
+    ))}
+  </ul>
 );
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -21,6 +21,7 @@ export const CommentList: React.FC<Props> = ({ comments }) => (
         commentName={comment.name}
         email={comment.email}
         commentBody={comment.body}
+        key={comment.id}
       />
     ))}
   </ul>
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,5 +1,58 @@
+// Libraries
 import React from 'react';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+// Styles
+import './PostInfo.scss';
+
+// Types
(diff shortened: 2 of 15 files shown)


src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,27 @@
...
+import { CommentInfo } from '../CommentInfo';
+
+type Props = {
+  comments: Comment[];
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <ul className="CommentList">
+    {comments.map((comment: Comment) => (
+      <CommentInfo
```
etojeDenys, line 20: Add a key prop to each CommentInfo component with a unique value, like the comment id.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,20 @@
...
 import React from 'react';
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
+// Styles
+import './UserInfo.scss';
+
+type Props = {
+  userName: string | undefined;
+  userEmail: string | undefined;
```
etojeDenys, line 9: Instead of using 'string | undefined', use 'string?' to mark optional properties in the Props type.

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,58 @@
...
+        {title}
+      </h3>
+
+      <UserInfo
+        userName={userName}
+        userEmail={userEmail}
+      />
+    </div>
+
+    <p className="PostInfo__body">
```
etojeDenys, line 42: Consider using a more descriptive classname for the component's elements, like 'post-info__body'.

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,58 @@
...
+
+// Component
+export const PostInfo: React.FC<Props> = ({
+  title,
+  userEmail,
+  userName,
+  body,
+  comments,
+}) => (
+  <li className="PostInfo">
```
etojeDenys, line 30: Consider using a more descriptive classname for the component's root element, like 'post-info'.

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,58 @@
...
+import { Comment } from '../../types/Comment';
+
+// Components
+import { UserInfo } from '../UserInfo';
+import { CommentList } from '../CommentList';
+
+type Props = {
+  title: string;
+  userEmail: string | undefined;
+  userName: string | undefined;
```
etojeDenys, line 17: It's better to use 'null' instead of 'undefined' for optional types in props.

src/components/UserInfo/UserInfo.tsx
etojeDenys: btw, is this prop really could be undefined?, and will your app not break if we pass undefined?

etojeDenys (changes requested):

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,39 @@
...
+
+// Components
+import { PostInfo } from '../PostInfo';
+
+// Functions
+function getComments(postId: number): Comment[] {
+  const foundComments
+    = commentsFromServer.filter(comment => comment.postId === postId);
+
+  return foundComments;
```
IvanFesenko, line 19:
```suggestion
  return commentsFromServer.filter(comment => comment.postId === postId);
```

IvanFesenko (approved):

PR #1903, opened 2023-06-23
React _static-list

ArtemKravcov85:
DEMO LINK

### DIFF
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,5 +1,17 @@
 import React from 'react';
+import { Comment } from '../../types/Comment';
+import { CommentInfo } from '../CommentInfo';
 
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
+import './CommentList.scss';
+
+type Props = {
+  comments: Comment[]
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <div className="CommentList">
+    {comments.map(comment => (
+      <CommentInfo comment={comment} />
+    ))}
+  </div>
 );
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,5 +1,39 @@
 import React from 'react';
+import { Post } from '../../types/Post';
+import { UserInfo } from '../UserInfo';
+import { CommentList } from '../CommentList';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
+import './PostInfo.scss';
+
+type Props = {
+  post: Post
+};
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <div className="PostInfo">
+    <div className="PostInfo__header">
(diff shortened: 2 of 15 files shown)


src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,17 @@
...
+import './CommentList.scss';
+
+type Props = {
+  comments: Comment[]
+};
+
+export const CommentList: React.FC<Props> = ({ comments }) => (
+  <div className="CommentList">
+    {comments.map(comment => (
+      <CommentInfo comment={comment} />
```
agniya-i, line 14: Add a unique `key` prop to the CommentInfo component to avoid issues when rendering the list.

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,15 @@
...
+type Props = {
+  posts: Post[];
+};
+
+export const PostList: React.FC<Props> = ({ posts }) => (
+  <div className="PostList">
+    { posts.map(post => (
+      <PostInfo post={post} />
+    )) }
+  </div>
```
agniya-i, line 14: Add a `key` prop here as well

src/types/User.ts
```diff
@@ -0,0 +1,6 @@
+export interface User {
+  id: number,
+  name: string,
+  username:string,
```
agniya-i, line 4:
```suggestion
  username: string,
```

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,39 @@
...
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <div className="PostInfo">
+    <div className="PostInfo__header">
+      <h3 className="PostInfo__title">{post.title}</h3>
+
+      <p>
+        {' Posted by  '}
+
+        <UserInfo user={post.user} />
```
agniya-i, line 20:
```suggestion
       {post.user && <UserInfo user={post.user} />}
```

It would be better to check if the user is not null in the parent component, not inside `UserInfo`, you can use `&&` operator

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,39 @@
...
+      </p>
+    </div>
+
+    <p className="PostInfo__body">
+      {post.body}
+    </p>
+
+    { post.comments.length
```
agniya-i, line 28:
```suggestion
    {post.comments.length
```

agniya-i (changes requested):

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,20 @@
...
-);
+import './UserInfo.scss';
+
+type Props = {
+  user: User | null
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  if (!user) {
+    return <></>;
```
witflash, line 12: If you want to return nothing inside your component you can use just `return null;`.
`null` is valid React element.

witflash (approved):
Well done! 👍🏻

PR #1904, opened 2023-06-23
React__static-list-of-posts

Soi4An:
DEMO - https://Soi4An.github.io/react_static-list-of-posts/

Test's problems:
  PostInfo:
      1) "before each" hook for "should have a correct title"
      2) should render no comments message is there are no comments
      3) should render another post info

...but all of these components are seen on the page

### DIFF
--- a/src/components/CommentInfo/CommentInfo.tsx
+++ b/src/components/CommentInfo/CommentInfo.tsx
@@ -1,5 +1,27 @@
 import React from 'react';
 
-export const CommentInfo: React.FC = () => (
-  <>Put the comment here</>
+import './CommentInfo.scss';
+
+import { Comment } from '../../types/comment';
+
+type Props = {
+  comment: Comment,
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => (
+  <div className="CommentInfo">
+    <div className="CommentInfo__title">
+      <strong className="CommentInfo__name">{comment.name}</strong>
+
+      {' by '}
+
+      <a className="CommentInfo__email" href={`mailto:${comment.email}`}>
+        {comment.email}
+      </a>
+    </div>
+
+    <div className="CommentInfo__body">
+      {comment.body}
+    </div>
+  </div>
 );
--- a/src/components/PostInfo/PostInfo.scss
+++ b/src/components/PostInfo/PostInfo.scss
@@ -1,1 +1,16 @@
-//
+.PostInfo {
+  margin: 10px auto;
+  width: 90%;
+  border: 1px solid #000;
+  border-radius: 8px;
+  background-color: lightgray;
+  padding: 1em;
+
+  &__title {
+    margin: 0;
+  }
+
+  &__header {
+    margin-bottom: 1em;
+  }
+}
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,5 +1,44 @@
 import React from 'react';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
-);
+import './PostInfo.scss';
+import { CommentList } from '../CommentList';
+import { UserInfo } from '../UserInfo';
+
+import { Comment } from '../../types/comment';
+import { User } from '../../types/user';
+import { Post } from '../../types/post';
+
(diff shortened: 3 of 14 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,44 @@
...
+
+  return (
+    <div className="PostInfo">
+      <div className="PostInfo__header">
+        <h3 className="PostInfo__title">{post.title}</h3>
+
+        <p>
+          {' Posted by  '}
+
+          <UserInfo user={users.find((u: User) => u.id === post.userId)} />
```
volodymyr-soltys97, line 29: Consider handling the case when the user is not found, as users.find() may return undefined. You can provide a default value or display a message.

src/components/PostInfo/PostInfo.scss
```diff
@@ -1 +1,16 @@
-//
+.PostInfo {
+  margin: 10px auto;
+  width: 90%;
+  border: 1px solid #000;
```
volodymyr-soltys97, line 4: Consider using a variable for the border color to maintain consistency and improve maintainability.

src/components/PostInfo/PostInfo.scss
```diff
@@ -1 +1,16 @@
...
+  border-radius: 8px;
+  background-color: lightgray;
```
volodymyr-soltys97, line 6: Consider using a variable for the background-color to maintain consistency and improve maintainability.

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,27 @@
...
-  <>Put the comment here</>
+import './CommentInfo.scss';
+
+import { Comment } from '../../types/comment';
+
+type Props = {
+  comment: Comment,
+};
+
+export const CommentInfo: React.FC<Props> = ({ comment }) => (
```
volodymyr-soltys97, line 11: It is recommended to use an interface instead of a type alias for component props to achieve better intellisense and code navigation experience.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,23 @@
...
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
-);
+import './UserInfo.scss';
+
+import { User } from '../../types/user';
+
+type Props = {
+  user: User | undefined,
```
volodymyr-soltys97, line 8: Change the type of the 'user' prop to be required instead of optional.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,23 @@
...
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  const userEmail: string = user ? user.email : '';
```
volodymyr-soltys97, line 12: Remove the conditional check for 'user' if the 'user' prop is required.

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,26 @@
...
+type Props = {
+  posts: Post[],
+  comments: Comment[],
+  users: User[],
+};
+
+export const PostList: React.FC<Props> = ({ posts, comments, users }) => (
+  <div className="PostList">
+    {posts.map((post: Post) => (
+      <PostInfo
```
volodymyr-soltys97, line 19: It is recommended to add a unique 'key' prop when mapping over an array to create a list of elements. This will help React to identify which items have changed, are added, or are removed, and will result in a better performance of the app.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/components/PostInfo/PostInfo.tsx
Soi4An: I'm doing these actions in UserInfo block

src/components/CommentInfo/CommentInfo.tsx
Soi4An: i used interface for start dates. I wrote these interfaces in folder types and they're imported in this type.

mykhalenych (approved):

PR #1912, opened 2023-06-27
solution

serhiilysenko:
https://serhiilysenko.github.io/react_static-list-of-posts

### DIFF
--- a/src/components/CommentInfo/CommentInfo.tsx
+++ b/src/components/CommentInfo/CommentInfo.tsx
@@ -1,5 +1,6 @@
 import { Comment } from '../../types/Comment';
 import './CommentInfo.scss';
+import '../CommentList/CommentList.scss';
 
 type Props = {
   comment: Comment;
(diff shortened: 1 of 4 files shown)


src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,35 @@
...
+    email,
+    body,
+  } = comment;
+
+  return (
+    <div className="CommentInfo">
+      <div className="CommentInfo__title">
+        <strong className="CommentInfo__name">{name}</strong>
+
+        {' by '}
```
VitaliyBodnar, line 20: it's not okay to add the spaces this way, better to use the margins and padding, for example:
`<span class='CommentInfo__title-by'>by</span>`

please check other places

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

PR #1916, opened 2023-07-19
Add posts

AmadeuszAndroid:
[DEMO LINK](https://amadeuszandroid.github.io/react_static-list-of-posts/)


src/App.tsx
```diff
@@ -2,108 +2,64 @@ import React from 'react';
...
+import postsFromServer from './api/posts';
+import commentsFromServer from './api/comments';
+import usersFromServer from './api/users';
+import { PostList } from './components/PostList';
+import { User } from './types/User';
+import { Comment } from './types/Comment';
+import { Post } from './types/Post';
+
+export const App: React.FC = () => {
+  const posts = postsFromServer.map(post => {
```
Radoslaw-Czerniawski, line 14: In functional components functions that are not dependent on state or props should be declared outside of the component or memoized with useCallback, otherwise when modifying the state they will be unnecessary redeclared lowering performance and potentially causing side-effects. 

This here implementation could also be written more functionally and much shorter:
image

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,31 @@
...
+import { Comment } from '../../types/Comment';
 
-export const CommentInfo: React.FC = () => (
-  <>Put the comment here</>
-);
+type CommentProps = {
+  comment: Comment;
+};
+
+export const CommentInfo: React.FC<CommentProps> = ({ comment }) => {
```
Radoslaw-Czerniawski, line 8: Generally mate-academy suggests using React.FC however I personally don't agree with that typing as it unnecessary adds children to props, this might create confusing components that shouldn't receive children (as they don't do anything with them) but can, more on that in this pr to official React repo 
-> https://github.com/facebook/create-react-app/pull/8177

src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,5 +1,31 @@
...
+  const { name, email, body } = comment;
```
Radoslaw-Czerniawski, line 9: This could also be destructured in the component's parameters on line 8:
image

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,44 @@
...
+import { User } from '../../types/User';
 
-export const PostInfo: React.FC = () => (
-  <>Put the post here</>
-);
+export type PostInfoProps = {
+  post: PostWithComments;
+};
+
+export type PostWithComments = {
```
Radoslaw-Czerniawski, line 11: This could be dependent on the Post type without the need for rewriting properties:
image

or with interface
image

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,44 @@
...
+  id: number;
+  user: User | undefined;
+  title: string;
+  body: string;
+  comments: Comment[];
+};
+
+export const PostInfo: React.FC<PostInfoProps> = ({ post }) => {
+  const {
```
Radoslaw-Czerniawski, line 20: Same here with destructuring

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,31 @@
...
 
-export const PostList: React.FC = () => (
-  <>Put the list here</>
-);
+export type PostListProps = {
+  posts: Post[],
+  comments: Comment[];
+};
+
+export const PostList: React.FC<PostListProps> = ({ posts, comments }) => {
```
Radoslaw-Czerniawski, line 10: This component should just receive Posts -> prepared posts with appropriate comments. Generally we are striving for Container -> Presentational component pattern (link [here](https://www.patterns.dev/posts/presentational-container-pattern)) where it should be App.tsx concern to prepare the data accordingly and this components only responsibility should be to display them. Functions that do the mapping in the example I showed you in App.tsx should do the trick

src/components/PostList/PostList.tsx
```diff
@@ -1,5 +1,31 @@
...
+  return (
+    <div className="PostList">
+      {posts.map(
+        post => {
```
Radoslaw-Czerniawski, line 14: There shouldn't be a need for this, correct comments should be attached to posts in the App.tsx and here only mapping of PostInfo should be happening

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,24 @@
 import React from 'react';
+import { User } from '../../types/User';
 
-export const UserInfo: React.FC = () => (
-  <>Put the user here</>
-);
+type UserProps = {
+  user: User | undefined;
```
Radoslaw-Czerniawski, line 5:
```tsx
type UserProps = {
  user?: User // same as above but shorter and making the user prop optional
}
```

Radoslaw-Czerniawski (changes requested):
Implement my suggestions, feel free to discuss them

src/types/Post.ts
```diff
@@ -0,0 +1,8 @@
+import { User } from './User';
+
+export interface Post {
+  user: User | undefined;
+  id: number;
+  title: string;
+  body: string;
+}
```
DorotaLeniecDev, line 8:
```suggestion
export interface Post {
  user: User | undefined;
  id: number;
  title: string;
  body: string;
  comments: Comment[],
}
```

src/types/Comment.ts
```diff
@@ -0,0 +1,9 @@
+import { Post } from './Post';
+
+export interface Comment {
+  post: Post | undefined;
```
DorotaLeniecDev, line 4:
```suggestion
  postid: number
```

DorotaLeniecDev (changes requested):
I see that you have not  responded to Radeks comments, check them and apply changes. You should pass posts with comments to the PostList component

src/components/CommentInfo/CommentInfo.tsx
AmadeuszAndroid: I added types in props parameters, because it is shorter.

src/components/CommentInfo/CommentInfo.tsx
AmadeuszAndroid: I moved destructuration to component's parameter.

src/App.tsx
AmadeuszAndroid: I extracted users and comments outside component.

src/components/PostInfo/PostInfo.tsx
AmadeuszAndroid: I destructured post.

src/components/PostList/PostList.tsx
AmadeuszAndroid: I removed comments from list.

src/components/PostList/PostList.tsx
AmadeuszAndroid: I added comments to post.

src/components/UserInfo/UserInfo.tsx
AmadeuszAndroid: I render user if exists in post.

src/types/Post.ts
AmadeuszAndroid: I added comments to post.

src/types/Comment.ts
AmadeuszAndroid: I deleted post from comment.

Radoslaw-Czerniawski (approved):
Looks good :)

PR #1936, opened 2023-08-24
Solution

Vira-V:
[DEMO LINK](https://Vira-V.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,15 +11,10 @@ import { Comment } from './types/Comment';
 import { Post } from './types/Post';
 import { PostList } from './components/PostList';
 
-function getUser(userId: number): User {
+function getUser(userId: number): User | null{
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -2,108 +2,45 @@ import React from 'react';
...
+
+function getUser(userId: number): User {
+  const findUser = usersFromServer.find(user => user.id === userId);
+
+  return findUser || {
+    id: 0,
+    name: '',
+    username: '',
+    email: '',
+  };
```
choeqq, line 22: user should be null as default - even when object is empty, it will render user info, if we do not have an user, we don't want to render anything related to user fields

choeqq (changes requested):

choeqq (approved):

PR #1941, opened 2023-08-26
Static list of posts solution

AndrzejPec:
[DEMO LINK](https://andrzejpec.github.io/react_static-list-of-posts/)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,5 +1,41 @@
...
+//         <strong>Post Title:</strong>
+//         {post.title}
+//       </div>
+//       <div>
+//         <strong>Post Body:</strong>
+//         {post.body}
+//       </div>
+//     </div>
+//   );
+// };
```
choeqq, line 41: please remove those comments

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,38 @@
...
+
+      <p className="PostInfo__body">
+        {post.body}
+      </p>
+
+      <hr />
+
+      {(post.comments !== null
+        && post.comments !== undefined
+        && post.comments.length > 0)
```
choeqq, line 31:
```suggestion
      {(post.comments.length > 0)
```
if comments would be null/undefined the value would be falsy, therefore b tag would be rendered, checking length is good enough

choeqq (changes requested):

src/components/PostInfo/PostInfo.tsx
AndrzejPec: possiblyNull

choeqq (approved):

PR #1945, opened 2023-08-27
Develop

MateuszSlezak:
https://MateuszSlezak.github.io/react_static-list-of-posts/

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,5 +1,35 @@
 import React from 'react';
+import { CommentList } from '../CommentList';
+import { UserInfo } from '../UserInfo/UserInfo';
+import { Post } from '../../types/Post';
+import './PostInfo.scss';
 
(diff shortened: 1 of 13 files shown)


src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,35 @@
...
+
+export const PostInfo: React.FC<Props> = ({ post }) => (
+  <div className="PostInfo">
+    <div className="PostInfo__header">
+      <h3 className="PostInfo__title">{post.title}</h3>
+
+      <p>
+        {' Posted by  '}
+
+        <UserInfo user={post.user} />
```
choeqq, line 19: user component has to be rendered, only when user is present

choeqq (changes requested):

src/components/PostInfo/PostInfo.tsx
MateuszSlezak: Done ;)

choeqq (approved):

PR #1966, opened 2023-11-30
Solution

Barosz30:
https://barosz30.github.io/react_static-list-of-posts/


src/App.tsx
```diff
@@ -2,108 +2,40 @@ import React from 'react';
...
+  const filteredcomment = commentsFromServer.filter(
+    comment => comment.postId === postId,
+  );
 
-        <p className="PostInfo__body">
-          deserunt eos nobis asperiores et hic est debitis repellat molestiae
-          optio nihil ratione ut eos beatae quibusdam distinctio maiores earum
-          voluptates et aut adipisci ea maiores voluptas maxime
-        </p>
+  return filteredcomment || [];
```
mosowiecka, line 26:
```suggestion
  return filteredcomment;
```
.filter() returns an empty array

mosowiecka (approved):

src/App.tsx
mbruhler, line 26: use camelCase for variables

mbruhler: That's right, you don't have to short circuit here

mbruhler (changes requested):
Fix your variables and conditional `|| []` and we're good to go

mbruhler (changes requested):
<del>Also</del>

> styles from App.scss should be moved to separate files, as well. E.g.: PostList.scss;

<del>and type files should be uppercase, you have `user.ts`</del>

it's not the case

mvjl000 (approved):
Looks good!

PR #1967, opened 2023-11-30
Solution

mosowiecka:
[DEMO LINK](https://mosowiecka.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentList/CommentList.tsx
+++ b/src/components/CommentList/CommentList.tsx
@@ -1,4 +1,5 @@
 // import React from 'react';
+import './CommentList.scss';
 import { Comment } from '../../types/Comment';
 import { CommentInfo } from '../CommentInfo';
 
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,4 +1,5 @@
 // import React from 'react';
+import './PostInfo.scss';
 import { Post } from '../../types/Post';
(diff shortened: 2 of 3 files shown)


src/components/CommentList/CommentList.tsx
```diff
@@ -1,5 +1,20 @@
...
-export const CommentList: React.FC = () => (
-  <>Put the list here</>
-);
+type Props = {
+  comments: Comment[],
+};
+
+export const CommentList = ({ comments }: Props) => {
+  if (!comments.length) {
+    return <b data-cy="NoCommentsMessage">No comments yet</b>;
```
adrian200002, line 11: Here <hr ></> is missing, look at original html.

adrian200002 (approved):
It looks like some css is missing too. I believe that the background of comments should not be white, referring to the readme, we had to move the styles from app.scss to separate files. E.g.: PostList.scss.

src/components/CommentList/CommentList.tsx
mbruhler: Yes, that's true. You can add it to top of `PostInfo` or you have to add it here

mbruhler (changes requested):
Your solution is still missing `import` in component files. Add them to properly style the page.

https://mosowiecka.github.io/react_static-list-of-posts/ looks like there is no CSS on the page

mbruhler (approved):

PR #1975, opened 2023-12-03
Solution

marcinzrodlowski:
[DEMO LINK](https://web3-sorcerer.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -11,7 +11,6 @@ export const PostInfo = ({
   post: {
     title,
     user,
-    userId,
     body,
     comments,
   },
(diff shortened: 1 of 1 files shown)


szymanski-marcin (approved):
looks good, appproved!

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,5 +1,42 @@
...
+    comments,
+  },
+} : Props) => (
+  <div className="PostInfo">
+    <div className="PostInfo__header">
+      <h3 className="PostInfo__title">
+        {title}
+      </h3>
+
+      {user && <UserInfo key={userId} user={user} />}
```
mvjl000, line 25: What's the purpose of `key` here? You are not mapping through any array

For Your Information: you can set `key` for non-list elements, but you have to do it consciously. By doing so, you can manually control when the component should re-render.

mvjl000 (changes requested):

PR #1988, opened 2024-03-05
solution

Arkadiuszx:
[DEMO LINK](https://Arkadiuszx.github.io/react_static-list-of-posts/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,9 @@ This task is similar to [Static List of TODOs](https://github.com/mate-academy/r
 - `Info` components should take one object and render its data.
 
 ## Instructions
+
 - Install Prettier Extention and use this [VSCode settings](https://mate-academy.github.io/fe-program/tools/vscode/settings.json) to enable format on save.
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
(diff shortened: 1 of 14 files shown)


Arkadiuszx:
The demo looks now the same, but I'm having issues with the tests.

yevhenii-pyl (approved):
```
  it('should have a link with mailto: user.email', () => {
    const user1 = {
      id: 1,
      name: 'Leanne Graham',
      username: 'Bret',
      email: 'Sincere@april.biz',
    };

    mount(<UserInfo user={user1} />);

    cy.get('.UserInfo').should('have.attr', 'href', 'mailto:Sincere@april.biz');
  });
```

Yep, noticed that thing with tests. They are expecting to receive the `user` object within the component. 

But hey, you are a developer here, give 'em hell, fill free to change it :)
`    mount(<UserInfo name={user1.name} email={user1.email} />); 
`

This is a minor adjustment to a testcase, but it makes sense due to your changes. 

In this case, taking test cases into consideration, I have misguided you a bit, however, I am still convinced that it's always better to pass separate props than the object into a component whenever it's possible (and also a good practice to check and modify tests if you will). 

Anyways, great job! Don't hesitate to change the world as you like it, no need to always play by the rules stated by others. 🥇

PR #2006, opened 2024-04-02
Solution

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/CommentInfo/CommentInfo.tsx
+++ b/src/components/CommentInfo/CommentInfo.tsx
@@ -1,3 +1,20 @@
 import React from 'react';
+import { Comment } from '../../types/Comment';
 
-export const CommentInfo: React.FC = () => <>Put the comment here</>;
+export const CommentInfo: React.FC<{ comment: Comment }> = ({ comment }) => {
+  return (
+    <div className="CommentInfo">
+      <div className="CommentInfo__title">
+        <strong className="CommentInfo__name">{comment.name}</strong>
+
+        {' by '}
+
+        <a className="CommentInfo__email" href={`mailto:${comment.email}`}>
+          {comment.email}
+        </a>
+      </div>
+
+      <div className="CommentInfo__body">{comment.body}</div>
+    </div>
+  );
+};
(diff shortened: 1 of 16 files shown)


src/components/CommentInfo/CommentInfo.tsx
```diff
@@ -1,3 +1,20 @@
 import React from 'react';
+import { Comment } from '../../types/Comment';
 
-export const CommentInfo: React.FC = () => <>Put the comment here</>;
+export const CommentInfo: React.FC<{ comment: Comment }> = ({ comment }) => {
```
yevhenii-pyl, line 4: It's always better to extract properties from a prop here on at the line below:

`{ comment: { name,  email ...} }`

src/components/CommentList/CommentList.tsx
```diff
@@ -1,3 +1,21 @@
 import React from 'react';
+import { Post } from '../../types/Post';
+import { CommentInfo } from '../CommentInfo/CommentInfo';
```
yevhenii-pyl, line 3: Can be simplified to `import { CommentInfo } from '../CommentInfo';`

(index file is always a default export area from the folder and by having such export from index 
`export * from './CommentInfo';` your code can easily be shorthanded :)

yevhenii-pyl (changes requested):
Good job here!
However, you have 3 failed tests. Kind of frustrating, isn't it? Your code is just fine! 
But if you look at `spec` file for `PostInfo` (where error messages are coming from) - you will see that your component supposed to receive a following structure as posts:

```
      const postWithComments = {
        userId: 2,
        id: 11,
        title: 'et ea vero quia laudantium autem',
        body: 'delectus reiciendis molestiae occaecati non minima eveniet qui voluptatibus\naccusamus in eum beatae sit\nvel qui neque voluptates ut commodi qui incidunt\nut animi commodi',
        user: user2,
        comments,  <== HERE
      };
```

See the difference? Comments. Not comment :) 
If you could fix this by adjusting your posts structure and corresponding type and we'll be good to go.

yevhenii-pyl (approved):
You got it! 🥇

PR #2013, opened 2024-07-09
My solution

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_static-list-of-posts/)

### DIFF
--- a/src/components/PostInfo/PostInfo.tsx
+++ b/src/components/PostInfo/PostInfo.tsx
@@ -1,3 +1,36 @@
 import React from 'react';
+import { Post } from '../../types/post';
+import { CommentList } from '../CommentList';
+import './PostInfo.scss';
+import { UserInfo } from '../UserInfo';
 
-export const PostInfo: React.FC = () => <>Put the post here</>;
+type Props = {
+  post: Post;
+};
+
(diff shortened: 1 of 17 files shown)


Karanelus:
What i need to do to pass the test, even when all the dependents are seen and work?

src/components/PostInfo/PostInfo.tsx
```diff
@@ -1,3 +1,36 @@
...
+const noComments: JSX.Element = (
+  <>
+    <hr />
+    <b data-cy="NoCommentsMessage">No comments yet</b>
+  </>
+);
+
+export const PostInfo: React.FC<Props> = ({ post }) => {
+  const { user, title, body, comments } = post;
+  const postetByText = `Posted by ${(<UserInfo user={user} />)}`;
```
natalia-klonowska, line 20: don't use template literals if your component returns jsx element and instead use conditional rendering for UserInfo so it only appears if user exists on post

natalia-klonowska (changes requested):

natalia-klonowska:
> What i need to do to pass the test, even when all the dependents are seen and work?

here is the problem that you need to fix
image

natalia-klonowska (approved):

